PR #3, opened 2017-03-05
Update various deps to latest versions.

burtonageo:


kvark:
@burtonageo thanks for the PR!
Sorry about it being left with no responses. The project has just been transferred under Gfx-rs organization, and I can guarantee you prompt reviews from now on. Speaking of the actual changes, I'm about to submit a PR doing this and much more, so closing.

PR #4, opened 2017-07-12
All around update

kvark:
r? @fkaa


fkaa:
LGTM! 👍

kvark:
@fkaa thanks for taking a look!
There is a problem I didn't manage to solve though - `window` example fails to find the vertex shader function. The shader library is binary, so I don't know what to do about it other than rewriting it. Perhaps, you could try getting it running on your end?

PR #6, opened 2017-07-13
Indirect argument buffers

kvark:


kvark:
Got clearance from @fkaa , proceeding to publish.

PR #7, opened 2017-07-18
Add MTLHeap bindings

JohnColanduoni:


src/heap.rs
```diff
@@ -0,0 +1,125 @@
+use {NSObjectPrototype, NSObjectProtocol, id};
```
kvark: please copy the license header in here

kvark:
Thank you! Published 0.4.2.

PR #9, opened 2017-09-16
Added setter for contentsScale

golddranks:
I need this setter to be able to set right scaling factor for the layer.

There's one but: I also added the definition for the `CGFloat` type. However, this type is already defined in `core-graphics` crate. It's just a one type, but if there is more redundancy, it should be considered to depend on `core-graphics`, since the types from there are used quite throughly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,11 @@ use std::any::Any;
 use std::fmt;
 use std::mem;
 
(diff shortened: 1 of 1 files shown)


golddranks:
Oops, sorry, I was careless. contentsScale is on CALayer, not CAMetalLayer.

PR #10, opened 2017-09-16
Added setter for contentsScale

golddranks:
I send a pull request a earlier today, but didn't verify that it actually worked :P Now I've done that. (The setter needed to be named `setContentsScale`, not `contentsScale`.)

I need this setter to be able to set right scaling factor for the layer.

There's one but: I also added the definition for the CGFloat type. However, this type is already defined in core-graphics crate. It's just a one type, but if there is more redundancy, it should be considered to depend on core-graphics, since the types from there are used quite throughly.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,11 @@ use std::any::Any;
 use std::fmt;
 use std::mem;
 
+#[cfg(target_pointer_width = "64")]
+pub type CGFloat = libc::c_double;
+#[cfg(not(target_pointer_width = "64"))]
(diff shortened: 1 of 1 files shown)


golddranks:
It's only a few lines of code, so I'd appreciate if someone could review this soonish. :)

kvark:
Sorry about the delay.
bors r+

golddranks:
Will Bors try again automatically or should something be done?

golddranks:
Great, thanks! Could you publish a new version, please? https://github.com/gfx-rs/metal-rs/issues/11

PR #13, opened 2017-10-04
Improve memory safety

JohnColanduoni:
This is very WIP, I just wanted to make sure this is the direction we want to go before I do the rest of the changes. The one file that is pretty much finished is `device.rs`.

As it stands `metal-rs` mostly leaves it to the user to determine lifetimes and remember to release objects. To rectify this we can use the [foreign-types](https://docs.rs/foreign-types/0.3.0/foreign_types/index.html) crate, which creates a `str`/`String`-type dichotomy for arbitrary C types. 

So far I'm following the convention that other ObjC crates like [core-graphics](https://github.com/servo/core-graphics-rs) do and using the naming convention `*mut MTLFoo` for the raw opaque pointer, `Foo` for the owned pointer, and `&FooRef` for the unowned pointer. Incoming arguments are always `&FooRef`, and return values are either `&FooRef` or `Foo` depending on the API's choices.

This isn't 100% safe because all Metal objects are potentially shared references, and there's the potential for nasty interactions with autorelease pools, but this at least gives us decent defaults.

### DIFF
--- a/examples/reflection/main.rs
+++ b/examples/reflection/main.rs
@@ -6,8 +6,12 @@
 // copied, modified, or distributed except according to those terms.
 
 extern crate metal_rs as metal;
+extern crate cocoa;
+#[macro_use] extern crate objc;
 
 use metal::*;
+
+use cocoa::foundation::NSAutoreleasePool;
 
 const PROGRAM: &'static str = "
     #include <metal_stdlib>\n\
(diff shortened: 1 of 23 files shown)


kvark:
Thank you, awesome stuff!
> As it stands metal-rs mostly leaves it to the user to determine lifetimes and remember to release objects. 

Is it not possible to automate?

JohnColanduoni:
Oh sorry I wasn't clear, that's what I'm fixing :)

kvark:
@JohnColanduoni is this ready to ship? The amount of code is insane, and given that I'm not a Metal/OSX expert, I wish we could get some help from @fkaa looking at it.

JohnColanduoni:
@kvark I think so. Most of the changes were just `self.0` -> `self` and dropping MTL prefixes in certain places. The main subtlety is whether I got the owned vs. unowned return types right (there were a couple of constructors returning unowned objects that tripped me up initially).

src/argument.rs
```diff
@@ -5,13 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use super::*;
```
kvark: I believe this is only idiomatic for test modules. Any reason not to pick the concrete imports here?

examples/reflection/main.rs
```diff
@@ -42,25 +46,33 @@ const PROGRAM: &'static str = "
...
         vertex_desc.serialize_descriptor(),
-        device,
+        &device,
         0x8,
         0x0
-    );
+    ) };
+
+    unsafe {
+        msg_send![pool, release];
```
kvark: why doesn't `NSAutoreleasePool` release itself automatically here?

src/heap.rs
```diff
@@ -5,128 +5,121 @@
...
         }
     }
 
     pub fn set_size(&self, size: NSUInteger) {
         unsafe {
-            msg_send![self.0, setSize: size];
+            msg_send![self, setSize: size];
         }
     }
 }
```
kvark: nit: newline, although it was in the old code

src/lib.rs
```diff
@@ -15,243 +15,242 @@ extern crate libc;
...
 }
 
-unsafe impl<T> objc::Message for id<T> { }
-
-#[allow(non_upper_case_globals)]
-pub const nil: id<()> = id(0 as *mut Object, PhantomData);
-
-pub trait AsObject {
-    fn as_obj(&self) -> *mut Object;
+macro_rules! try_objc {
```
kvark: neat!

src/lib.rs
```diff
@@ -346,3 +333,12 @@ pub use vertexdescriptor::*;
...
 
+#[inline]
+unsafe fn obj_drop<T>(p: *mut T) {
+    msg_send![(p as *mut Object), release];
+}
+
+#[inline]
+unsafe fn obj_clone<T: 'static>(p: *mut T) -> *mut T {
+    msg_send![(p as *mut Object), retain]
+}
```
kvark: nit: newline

src/argument.rs
JohnColanduoni: Laziness? :P

examples/reflection/main.rs
JohnColanduoni: Any releasing of an `NSAutoreleasePool` can invalidate certain unowned references that the Metal API produces, so that call can cause memory unsafety if executed at the wrong time. The `NSAutoreleasePool` binding is also from another crate (`cocoa` IIRC) so I'd have to duplicate it to add that functionality.

JohnColanduoni:
@kvark Yep, none of those should be necessary after we update the `gfx` backend.

kvark:
Excellent!
bors r+

kvark:
last time...
bors r+

notriddle:
https://github.com/gfx-rs/metal-rs/pull/14

kvark:
Thanks @notriddle ! (wow, how do you know where to show up?)
I've just given up on waiting for CI.

notriddle:
Easy; if one PR gets an excessive number of bors comments, I look at it.

PR #14, opened 2017-10-18
Create bors.toml with five-hour timeout

notriddle:


kvark:
Thank you!
(to self: we still need to exclude staging.tmp from CI configs...)
bors r+

PR #16, opened 2017-10-22
Dependency updates, argument buffer API fix

kvark:
Also bumps the version to 0.5.0 ready for publishing

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -52,9 +52,9 @@ pub enum MTLArgumentBuffersTier {
 
 bitflags! {
     struct MTLPipelineOption: NSUInteger {
-        const MTLPipelineOptionNone           = 0;
-        const MTLPipelineOptionArgumentInfo   = 1 << 0;
-        const MTLPipelineOptionBufferTypeInfo = 1 << 1;
+        const None           = 0;
+        const ArgumentInfo   = 1 << 0;
+        const BufferTypeInfo = 1 << 1;
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/device.rs
```diff
@@ -277,7 +277,7 @@ impl DeviceRef {
         }
     }
 
-    pub fn new_argument_encoder(&self, arguments: &Array<ArgumentDescriptor>) -> ArgumentEncoder {
```
JohnColanduoni: I think this should be `&ArrayRef<ArgumentDescriptor>`, since there's no need to consume a retain count here. You're right that it definitely shouldn't be `&Array` though, unfortunately `msg_send!` will implicitly accept just about anything argument-wise :/

kvark:
It depends on bitflags being public or not...

kvark:
@JohnColanduoni I think technically we should either yank 0.5.0 (is anyone other than gfx-rs using it?), or publish the new code as 0.6.0 (perhaps, merged with the other open PR).

Also, I haven't found a way to create `ArrayRef`. Would you mind following up with the way you think would be correct there?

JohnColanduoni:
@kvark Yeah that's my bad, I couldn't use the `foreign_types` macro to generate the trait implementations for `Array` because of the generic parameter and I forgot to make it Deref properly. I'll push a quick fix if you don't mind.

kvark:
Sure, go for it.

kvark:
@JohnColanduoni want to make a PR with those changes?

JohnColanduoni:
@kvark Sure: https://github.com/gfx-rs/metal-rs/pull/19

kvark:
Thanks!
replaced by #19

PR #18, opened 2017-10-22
Add .DS_Store to .gitignore

burtonageo:


JohnColanduoni:
I'd generally recommend you use a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for that on OSX but there's no reason not to add it here too.

kvark:
I agree, it dosen't make much sense to have it in all the github projects...

PR #19, opened 2017-10-24
Argument buffer fix

JohnColanduoni:
A continuation of https://github.com/gfx-rs/metal-rs/pull/16


kvark:
@JohnColanduoni looks like you forgot to update the argument buffer example code

PR #20, opened 2017-10-27
Dual-source blending

kvark:


kvark:
Should be harmless.
bors r+

PR #21, opened 2017-12-04
Add function constants

grovesNL:
Add initial support for `MTLFunctionConstantValues` to use in gfx specialization constants.

Please take a look and let me know if this seems right. I'm not sure how the deallocations are supposed to work, but I've tried to follow the other examples.

Note this changes the signature of `get_function`. I can introduce a separate function instead if that's preferable.

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -9,7 +9,9 @@ use super::*;
 
 use objc::runtime::{Class, Object, YES, NO};
 use objc_foundation::{NSString, INSString, NSArray};
+use cocoa::foundation::{NSUInteger};
 use foreign_types::ForeignType;
+use std::ffi::CStr;
 
 pub enum MTLVertexAttribute {}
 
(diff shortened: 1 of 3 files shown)


kvark:
@JohnColanduoni would you be able to take a look? Otherwise I'll do, but I don't have as much insight on the memory management side of things.

src/library.rs
```diff
@@ -195,18 +219,22 @@ impl LibraryRef {
...
             use cocoa::foundation::NSString as cocoa_NSString;
             use cocoa::base::nil as cocoa_nil;
 
             let nsname = cocoa_NSString::alloc(cocoa_nil).init_str(name);
-            //let nsname = NSString::from_str(name);
-            let function: *mut MTLFunction = msg_send![self, newFunctionWithName:nsname];
+            
+            let function: *mut MTLFunction = match constants {
+                Some(c) => try_objc!{ err => msg_send![self, newFunctionWithName:nsname constantValues:c error:&mut err] },
+                None => try_objc!{ err => msg_send![self, newFunctionWithName:nsname] }
```
JohnColanduoni: ‘try_objc’ shouldn’t be used for this match case because there is no ‘err’ parameter to the method.

src/library.rs
```diff
@@ -195,18 +219,22 @@ impl LibraryRef {
...
+            };
+
             if !function.is_null() {
-                Some(Function::from_ptr(function))
+                Ok(Function::from_ptr(function))
             } else {
-                None
+                Err("Unhandled".to_owned())
```
JohnColanduoni: This case will only be triggered if the function does not exist so we can put that as the error message.

src/library.rs
grovesNL: Good catch, thanks

grovesNL:
Thanks for the review! I updated and squashed my commits.

PR #27, opened 2018-01-24
Dispatch and more compute states

kvark:


kvark:
bors r=glrovesNL

kvark:
Published in https://crates.io/crates/metal-rs/0.6.4
Travis OSX is not in the mood lately...

src/encoder.rs
```diff
@@ -434,9 +438,41 @@ impl ComputeCommandEncoderRef {
         }
     }
 
+    pub fn set_texture(&self, index: u64, texture: Option<&TextureRef>) {
```
JohnColanduoni: The u64 should be an NSUInteger.

src/encoder.rs
```diff
@@ -434,9 +438,41 @@ impl ComputeCommandEncoderRef {
...
+        unsafe {
+            msg_send![self, setTexture:texture
+                               atIndex:index]
+        }
+    }
+
+    pub fn set_sampler_state(&self, index: u64, sampler: Option<&SamplerStateRef>) {
```
JohnColanduoni: Same here

kvark:
Thanks @JohnColanduoni ! The notes are addressed in #28

PR #32, opened 2018-02-27
Add missing constants from MTLPixelFormat, make it Eq

### LINKED ISSUE #30: MTLPixelFormat does not implement PartialEq
`MTLPixelFormat` should derive `PartialEq`, or really `Eq`, since it's just a plain enum.
Most of the time it's going to be used with `match`, but having it implement `PartialEq` lets us do useful things like
```
debug_assert_eq!(src.raw.pixel_format(), dst.raw.pixel_format());
```

### LINKED ISSUE #31: MTLPixelFormat has missing values
`MTLPixelFormat` has values which are missing (at least as of macOS 10.13).

There are also some formats which are available on iOS only which I'm assuming were omitted on purpose.

These are the missing formats (copied from MTLPixelFormat.h):
```
MTLPixelFormatBGR10A2Unorm  NS_AVAILABLE(10_13, 11_0) = 94,

MTLPixelFormatDepth16Unorm          NS_AVAILABLE_MAC(10_12) = 250,

MTLPixelFormatX32_Stencil8  NS_AVAILABLE(10_12, 10_0) = 261,
MTLPixelFormatX24_Stencil8  NS_AVAILABLE_MAC(10_12) = 262,
```

mjadczak:
Fixes #30 and fixes #31

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -7,7 +7,7 @@
 
 #[repr(u64)]
 #[allow(non_camel_case_types)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum MTLPixelFormat {
    Invalid                = 0,
    A8Unorm                = 1,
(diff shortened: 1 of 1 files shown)


kvark:
Published in https://crates.io/crates/metal-rs/0.6.6
Technically, a breaking change, but... hopefully no one is affected.

PR #33, opened 2018-02-27
Add device registryID

keringar:
I don't have access to a MacOS device right now, but this should compile and run fine. If you would like to wait a couple days I might be able to get access to an Apple device to test this on. More documentation found here https://developer.apple.com/documentation/metal/mtldevice/2915737-registryid

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -132,6 +132,12 @@ impl DeviceRef {
         }
     }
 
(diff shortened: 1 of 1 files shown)


kvark:
Published in https://crates.io/crates/metal-rs/0.6.6

PR #34, opened 2018-03-01
Argument encoding cleanup

kvark:
Unfortunately, a breaking, since the old `set_buffers` appear to be incorrect (please confirm!).


src/encoder.rs
```diff
@@ -513,55 +513,91 @@ impl ArgumentEncoderRef {
...
+                atIndex: at_index
+            ]
+        }
+    }
+
+    pub fn set_buffers(&self, data: &[&BufferRef], offsets: &[NSUInteger], start_index: NSUInteger) {
+        unsafe {
+            msg_send![self,
+                setBuffers: data.as_ptr()
+                offsets: offsets
```
mjadczak: Shouldn't `offsets` also be converted into a pointer, or is this implicit?
The obj C method takes `offsets` to be `const NSUInteger *`

src/encoder.rs
```diff
@@ -513,55 +513,91 @@ impl ArgumentEncoderRef {
...
-                              atIndex:at_index]
+            msg_send![self,
+                setBuffer: buffer
+                offset: offset
+                atIndex: at_index
+            ]
+        }
+    }
+
+    pub fn set_buffers(&self, data: &[&BufferRef], offsets: &[NSUInteger], start_index: NSUInteger) {
```
mjadczak: Could also use e.g. `RangeFrom` to be more semantic about `start_index` (also elsewhere).

src/encoder.rs
```diff
@@ -513,55 +513,91 @@ impl ArgumentEncoderRef {
...
+        unsafe {
+            msg_send![self,
+                setBuffers: data.as_ptr()
```
mjadczak: Do we want to assert that `offsets.len() <= data.len()`?

src/encoder.rs
kvark: It has neat semantics, but a little bit too much sugar for this kind of library.

kvark:
@mjadczak great points spotted, thank you for the review!
bors r=mjadczak

PR #37, opened 2018-04-17
More pixel formats

kvark:
r? @grovesNL 
Note: as with previous PR, this is technically a breaking change, since we are extending an enum. However, we don't expect anyone to it exhaustively. If any complaints come, we'll yank it in no time.


grovesNL:
From quickly glancing through it we could also add

```
R8Unorm_sRGB = 11,
RG8Unorm_sRGB = 31,
```

or we can update in a later PR

kvark:
Nicely spotted! PTAL

kvark:
Bors delegate=grovesNL

grovesNL:
👍 
bors r+

PR #39, opened 2018-05-02
Copy texture to texture

kvark:


kvark:
@grovesNL jeez, I'm being super sloppy. Thanks for noticing this! PTAL

kvark:
I believe it's good now :crossed_fingers: 
bors r=grovesNL

grovesNL:
It was worth trying 😄 you should be able to merge now at least

kvark:
@grovesNL sorry about that! Please try again

PR #41, opened 2018-05-04
Add more vertex formats

grovesNL:
Add some newer vertex formats


kvark:
Similarly to some of the previous PRs, this is technically a breaking change, but we close our eyes on it (unless complained).

PR #46, opened 2018-05-08
Version bump and cocoa dependency update

kvark:


kvark:
This is fairly trivial.
bors r+

PR #51, opened 2018-05-14
Fix bitflags to use namespaces properly, add blit options

kvark:


src/pipeline/render.rs
```diff
@@ -51,12 +51,11 @@ pub enum MTLBlendOperation {
...
-        const MTLColorWriteMaskRed   = 0x1 << 3;
-        const MTLColorWriteMaskGreen = 0x1 << 2;
-        const MTLColorWriteMaskBlue  = 0x1 << 1;
-        const MTLColorWriteMaskAlpha = 0x1 << 0;
-        const MTLColorWriteMaskAll   = 0xf;
+        const Red   = 0x1 << 3;
+        const Green = 0x1 << 2;
+        const Blue  = 0x1 << 1;
+        const Alpha = 0x1 << 0;
+        const All   = 0xf;
```
grovesNL: Should we keep `All` if we're dropping `None` (bitflags already has `all` and `empty`?

kvark:
Thanks for the review!
Removed that constant now. Looking forward for Rust constant expressions to cover this ;)

Bors r=grovesNL

PR #53, opened 2018-05-25
Indirect draw calls

kvark:


kvark:
Yes, we shouldn't change the old method, but mark it deprecated for now.

kvark:
Alright, made it non-breaking for now.
Bors r=grovesNL

PR #57, opened 2018-06-14
[WIP] Experimental ObjC selector hack

kvark:
Fixes the slow-ness of our ObjC messages.


kvark:
Addressed upstream in rust-objc

PR #59, opened 2018-06-29
Fix the zombie error of shader compilation

kvark:
What I think is happening is that the library object keeps a reference to the error, so we shouldn't release it if the library is created.


kvark:
Bors r=grovesNL

PR #60, opened 2018-07-10
Fix compilation for iOS with cargo lipo

Michael-Lfx:


Cargo.toml
```diff
@@ -10,6 +10,9 @@ readme = "README.md"
 keywords = ["metal", "graphics", "bindings"]
 license = "MIT OR Apache-2.0"
 
+[lib]
+crate-type = ["staticlib", "rlib"]
```
grovesNL: Why is this required?

src/lib.rs
```diff
@@ -28,7 +28,17 @@ use std::ops::Deref;
 use std::borrow::{Borrow, ToOwned};
 
 use objc::runtime::{Object, Class, YES, NO};
+#[cfg(target_os = "macos")]
 use cocoa::foundation::NSSize;
+#[cfg(target_os = "ios")]
+#[repr(C)]
+#[derive(Debug, Clone)]
+pub struct CGSize {
```
grovesNL: Do we need to expose `CGSize` here? This seems problematic if other crates start to use this type directly (ideally this could be defined upstream like [`CGPoint`/`NSPoint`](https://github.com/servo/core-foundation-rs/blob/master/core-graphics/src/geometry.rs#L37))

Cargo.toml
Michael-Lfx: Because I use `cargo lipo` to build library for iOS. If there is no `[lib]` session in toml, `cargo lipo` will exit with 
> Failed to parse cargo read-manifest output: No target with kind staticlib found.

Hence, `staticlib` is required by `cargo lipo`. Why use "rlib"? If there is another Rust project use the project as extern crate on iOS, lacking of "rlib" results in 
> Failed to find extern crate XXXXX

With above code, `cargo lipo` works for my team.

src/lib.rs
Michael-Lfx: When building for iOS, it come out a error just like "no CGSize define". So, Add a definition here to fix the error.

Cargo.toml
grovesNL: Is it possible to do this only for Android and iOS targets? Something like https://github.com/mozilla/toodle/pull/10#discussion_r159790273

src/lib.rs
grovesNL: So it seems like [cocoa already exposes `CGSize`](https://github.com/servo/core-foundation-rs/blob/master/core-graphics/src/geometry.rs#L37), could we just use that instead?

([`CGSize`](https://github.com/servo/core-foundation-rs/blob/master/core-graphics/src/geometry.rs#L37) and [`NSSize`](https://github.com/servo/core-foundation-rs/blob/master/cocoa/src/foundation.rs#L67) are defined separately in cocoa)

Michael-Lfx: I just look the source of `core_graphics` inside `cocoa`. Tuns out that it just provide public access to macOS, not iOS included. Here are the code.

```
#[cfg(target_os = "macos")]
pub mod event_source;
pub mod font;
pub mod geometry;
```

`CGSize` is defined in mod `geometry`. Should I submit a PR to project [core-foundation-rs](https://github.com/servo/core-foundation-rs)?

Michael-Lfx: Turns out I need to modify `cocoa` to make pub access to crate core_graphics. And `metal-rs/src/lib.rs` still need the fellowing code for iOS compilation.

```
#[cfg(target_os = "macos")]
use cocoa::foundation::NSSize;
#[cfg(target_os = "ios")]
use cocoa::core_graphics::geometry::CGSize;
#[cfg(target_os = "ios")]
type NSSize = CGSize;
```

Cargo.toml
Michael-Lfx: `cargo lipo` failed to parse the way [mozilla/toodle#10 (comment)](https://github.com/mozilla/toodle/pull/10#discussion_r159790273) discussed. Tested with the following code. 

```
[lib]
name = "metal_rs"
[target.'cfg(any(target_os="android", target_os="ios"))'.lib]
crate-type = ["staticlib", "rlib"]
```

src/lib.rs
grovesNL: Couldn't we just use `core_graphics::geometry::CGSize` directly instead of re-exporting it through cocoa?

It seems like [`CGSize` is actually the correct type here](https://developer.apple.com/documentation/quartzcore/cametallayer/1478174-drawablesize?language=objc) anyway, so maybe we don't need `NSSize` at all.

Cargo.toml
TimNN: The functionality offered by `cargo-lipo` only makes sense if a `staticlib` built. If there is a way to tell `cargo build` that you want a static lib, I'd gladly remove that check. (If having something like `#![cfg_attr(target_os = "iOS"), crate_type = "staticlib"]` in `lib.rs` works, I'd be happy to accept a PR to `cargo-lipo` that conditionally disables the check of the manifest).

---

On a separate note, I don't think it makes sense to build this crate with `cargo-lipo`: This crate only provides a Rust API, as far as I understand. `cargo-lipo` only really works if you're crate exposes a C-API, which you can call from you Objective-C / Swift code.

grovesNL: Thanks for the clarification @TimNN. It seems to make sense for `cargo-lipo` to force the `staticlib` requirement in that case.

@Michael-Lfx what's the reason you are building this crate directly with `cargo-lipo` (see the note about crates providing C externs above)?

kvark:
Overshadowed by #69

PR #62, opened 2018-07-21
Add feature set queries

AIOOB:
I thought I'd check that this is a sensible approach before I finished going through the feature tables as it's a bit tedious. As far as I can tell there is no nice source of this information bar the feature table pdf, I would love to know if anyone has found any other sources.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -44,10 +44,1230 @@ pub enum MTLFeatureSet {
     tvOS_GPUFamily2_v1 = 30003,
     macOS_GPUFamily1_v1 = 10000,
     macOS_GPUFamily1_v2 = 10001,
-    macOS_ReadWriteTextureTier2 = 10002,
+    //macOS_ReadWriteTextureTier2 = 10002, TODO: Uncomment when feature tables updated
     macOS_GPUFamily1_v3 = 10003,
-    macOS_GPUFamily1_v4 = 10004,
-    macOS_GPUFamily2_v1 = 10005,
+    //macOS_GPUFamily1_v4 = 10004,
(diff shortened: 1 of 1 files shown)


src/device.rs
```diff
@@ -44,10 +44,1299 @@ pub enum MTLFeatureSet {
...
+    //macOS_ReadWriteTextureTier2 = 10002, TODO: Uncomment when feature tables updated
     macOS_GPUFamily1_v3 = 10003,
-    macOS_GPUFamily1_v4 = 10004,
-    macOS_GPUFamily2_v1 = 10005,
+    //macOS_GPUFamily1_v4 = 10004,
+    //macOS_GPUFamily2_v1 = 10005,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct PixelFormatCapabilities {
```
kvark: let's make at a bit set?

src/device.rs
```diff
@@ -44,10 +44,1299 @@ pub enum MTLFeatureSet {
...
+            blend: false,
+            msaa: false,
+            resolve: false,
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum OS {
+    iOS, tvOS, macOS
```
kvark: let's add `watchOS` here

AIOOB: I've kept this private so that it can easily map to the feature set tables, if I add watchOS then it would require a load of weird cases in all the functions. Also if you would like this to be public I think a different structure might be sensible where the version comes along with it.

PR #63, opened 2018-07-27
Get all devices

kvark:


src/device.rs
```diff
@@ -68,6 +68,7 @@ bitflags! {
 #[link(name = "Metal", kind = "framework")]
 extern {
     fn MTLCreateSystemDefaultDevice() -> *mut MTLDevice;
+    fn MTLCopyAllDevices() -> *mut Object; //TODO: Array
```
grovesNL: We already include `NSArray` from foundation, do we need something else here?

kvark:
It doesn't work because of an undefined layout. I figured we can fix it later since it's private API anyway.

grovesNL:
Ok sounds good
bors r+

kvark:
https://crates.io/crates/metal-rs/0.10.4

PR #68, opened 2018-08-09
Fix set_fragment_textures

kvark:
This is a follow-up to #67


kvark:
This is straightforward and tested.
bors r=common sense

PR #69, opened 2018-08-10
fix iOS: use CGSize instead of NSSize. this is the correct type anyway

mtak-:
This is based on the work in this PR #60 by @Michael-Lfx which has not yet been accepted. I am guessing it's abandoned? If not feel free to reject this.

I, too, want this feature merged to help get `metal` (and ultimately `gfx`) building on iOS.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@
 #![allow(non_upper_case_globals)]
 
 extern crate cocoa;
+extern crate core_graphics;
 #[macro_use]
 extern crate bitflags;
 extern crate libc;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -28,14 +29,10 @@ use std::ops::Deref;
 use std::borrow::{Borrow, ToOwned};
 
 use objc::runtime::{Object, YES, NO};
-use cocoa::foundation::NSSize;
+pub use core_graphics::base::CGFloat;
```
grovesNL: @kvark do you have any strong preference about reexporting types?

kvark:
Need to check the backend use of core_graphics. If it needs to depend on it directly anyway, then I'd rather not have metal-rs re-exporting those types.

mtak-:
I've removed the reexports. gfx-metal-backend only needs `CGRect` and `CGFloat` from core_graphics. `CGRect` is not used by metal-rs so gfx-metal-backend would need core_graphics anyway.

kvark:
Bors r+

kvark:
@mtak- we'll publish 0.12 once #70 is merged

PR #71, opened 2018-08-20
Add fill_buffer and setting the visibility results

kvark:


src/renderpass.rs
```diff
@@ -331,6 +331,12 @@ impl RenderPassDescriptorRef {
         }
     }
 
+    pub fn set_visibility_result_buffer(&self, buffer: &BufferRef) {
```
grovesNL: We might want this to be `Option<&BufferRef>` instead

Issue #52, opened 2018-05-22
Determine supported functions


grovesNL:
See https://github.com/gfx-rs/gfx/issues/2058#issuecomment-390864184

We need a way to determine whether certain functions are supported, which depends on the device OS and Metal version.

Ideally metal-rs could provide some way to query this information (i.e. `fn supports_xyz() -> bool`) or guard against these cases (i.e. return a certain `Err` type), instead of callers having to perform their own version checks and maintain feature tables for each version.

kvark:
If we can get away with just a standard feature discovery without double checking each function for existence, that would be ideal for me. Treating every entry point as optional doesn't appear to be an elegant path forward.

grovesNL:
Opened a bug report with Apple to clarify. Hopefully they will be able to provide some guidance for the best approach here.

PR #73, opened 2018-08-31
[iOS] audit msg_send! for ios/tvos compatibility

mtak-:
While working on getting the gfx/quad example going, I ran into some metal calls that are unsupported on iOS. So I did a pass on all the ffi for os compatibility, and fixed things where I could.

- metal-rs is using some private APIs, and apple has a history of rejecting apps from the app store for doing that
- there's no managed memory on iOS
- there's no layered rendering on iOS

- iOS 8.0 is missing some features from osx 10.11
- iOS 9.0 is roughly equivalent in metal support to osx 10.11
- iOS 10.0 is roughly equivalent in metal support to osx 10.12
- iOS 11.0 is roughly equivalent in metal support to osx 10.13

I added a bunch of comments above things that were unsupported on various os versions. Should I leave those in or nuke'em?

This breaks gfx-rs on iOS. I'll make a PR there shortly as well.

### DIFF
--- a/examples/caps/main.rs
+++ b/examples/caps/main.rs
@@ -12,11 +12,17 @@ use metal::*;
 fn main() {
     let device = Device::system_default();
 
-    println!("Vendor: {:?}", device.vendor());
-    println!("Family: {:?}", device.family_name());
+    #[cfg(feature = "private")]
+    {
+        println!("Vendor: {:?}", unsafe { device.vendor() });
+        println!("Family: {:?}", unsafe { device.family_name() });
+    }
     println!("Max threads per threadgroup: {:?}", device.max_threads_per_threadgroup());
-    println!("Integrated GPU: {:?}", device.is_low_power());
-    println!("Headless: {:?}", device.is_headless());
-    println!("D24S8: {:?}", device.d24_s8_supported());
+    #[cfg(target_os = "macos")]
+    {
+        println!("Integrated GPU: {:?}", device.is_low_power());
+        println!("Headless: {:?}", device.is_headless());
+        println!("D24S8: {:?}", device.d24_s8_supported());
+    }
     println!("Indirect argument buffer: {:?}", device.argument_buffers_support());
 }
--- a/examples/compute/compute-argument-buffer.rs
+++ b/examples/compute/compute-argument-buffer.rs
@@ -62,7 +62,10 @@ fn main() {
     let pipeline_state_descriptor = ComputePipelineDescriptor::new();
     pipeline_state_descriptor.set_compute_function(Some(&kernel));
 
-    let pipeline_state = device.new_compute_pipeline_state(&pipeline_state_descriptor).unwrap();
+    let pipeline_state = device
+        .new_compute_pipeline_state_with_function(
+            pipeline_state_descriptor.compute_function().unwrap(),
+        ).unwrap();
 
     encoder.set_compute_pipeline_state(&pipeline_state);
     encoder.set_buffer(0, Some(&arg_buffer), 0);
--- a/src/device.rs
+++ b/src/device.rs
@@ -36,10 +36,18 @@ pub enum MTLFeatureSet {
     iOS_GPUFamily2_v4 = 9,
     iOS_GPUFamily3_v3 = 10,
     iOS_GPUFamily4_v1 = 11,
+    // iOS_GPUFamily1_v5 = 12, TODO: Uncomment when feature tables updated
+    // iOS_GPUFamily2_v5 = 13,
+    // iOS_GPUFamily3_v4 = 14,
+    // iOS_GPUFamily4_v2 = 15,
+    
     tvOS_GPUFamily1_v1 = 30000,
     tvOS_GPUFamily1_v2 = 30001,
     tvOS_GPUFamily1_v3 = 30002,
     tvOS_GPUFamily2_v1 = 30003,
+    // tvOS_GPUFamily1_v4 = 30004,
+    // tvOS_GPUFamily2_v2 = 30005,
+
     macOS_GPUFamily1_v1 = 10000,
     macOS_GPUFamily1_v2 = 10001,
     //macOS_ReadWriteTextureTier2 = 10002, TODO: Uncomment when feature tables updated
@@ -1355,18 +1363,16 @@ impl DeviceRef {
         }
     }
 
-    pub fn vendor(&self) -> &str {
-        unsafe {
-            let name: &NSString = msg_send![self, vendorName];
-            name.as_str()
-        }
+    #[cfg(feature = "private")]
+    pub unsafe fn vendor(&self) -> &str {
+        let name: &NSString = msg_send![self, vendorName];
+        name.as_str()
     }
 
-    pub fn family_name(&self) -> &str {
-        unsafe {
-            let name: &NSString = msg_send![self, familyName];
-            name.as_str()
-        }
+    #[cfg(feature = "private")]
+    pub unsafe fn family_name(&self) -> &str {
+        let name: &NSString = msg_send![self, familyName];
+        name.as_str()
     }
 
     pub fn registry_id(&self) -> u64 {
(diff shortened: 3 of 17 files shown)


src/device.rs
```diff
@@ -1355,13 +1355,15 @@ impl DeviceRef {
         }
     }
 
+    // TODO: private API
```
kvark: nicely spotted!
fortunately, we are't appear to be using this. How about adding a `private` feature that is not enabled by default, allowing to expose functions like this?

src/device.rs
```diff
@@ -1377,27 +1379,44 @@ impl DeviceRef {
 
     pub fn max_threads_per_threadgroup(&self) -> MTLSize {
         unsafe {
-            msg_send![self, maxThreadsPerThreadgroup]
+            if cfg!(target_os = "ios") {
+                let has_max_threads_per_threadgroup: BOOL = msg_send![self, respondsToSelector:sel!(maxThreadsPerThreadgroup)];
+                match has_max_threads_per_threadgroup {
+                    YES => msg_send![self, maxThreadsPerThreadgroup],
+                    NO => MTLSize { width: 512, height: 512, depth: 512 },
```
kvark: Where do those numbers come from?

src/device.rs
```diff
@@ -1377,27 +1379,44 @@ impl DeviceRef {
...
         }
     }
 
     pub fn is_low_power(&self) -> bool {
-        unsafe {
-            match msg_send![self, isLowPower] {
-                YES => true,
-                NO => false,
-                _ => unreachable!()
+        if cfg!(target_os = "macos") {
```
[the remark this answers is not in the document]
kvark: I don't think `metal-rs` should be making this decision though. It's really meant to be a low-level wrapper, so that we know 1 method corresponds to exactly 1 message in 99% of cases.

I am open to other opinions though, especially one of @grovesNL  :)

src/device.rs
```diff
@@ -1377,27 +1379,44 @@ impl DeviceRef {
 
     pub fn max_threads_per_threadgroup(&self) -> MTLSize {
         unsafe {
-            msg_send![self, maxThreadsPerThreadgroup]
+            if cfg!(target_os = "ios") {
+                let has_max_threads_per_threadgroup: BOOL = msg_send![self, respondsToSelector:sel!(maxThreadsPerThreadgroup)];
```
kvark: nit: can we `match msg_send![..]` right away?

src/heap.rs
```diff
@@ -50,7 +51,7 @@ impl HeapRef {
 
     pub fn max_available_size(&self, alignment: NSUInteger) -> NSUInteger {
```
kvark: I do wonder though if `maxAvailableSize` was deprecated at some point... great find btw!

src/library.rs
```diff
@@ -180,19 +183,41 @@ impl CompileOptionsRef {
 
     pub fn language_version(&self) -> MTLLanguageVersion {
         unsafe {
-            msg_send![self, languageVersion]
+            if cfg!(not(target_os = "ios")) {
+                msg_send![self, languageVersion]
+            } else {
+                let has_language_version: BOOL = msg_send![self, respondsToSelector:sel!(languageVersion)];
+                match has_language_version {
```
kvark: looking at this logic, it is indeed tempting to make metal-rs to abstract over macOS/tvOS/iOS at least...

src/resource.rs
```diff
@@ -31,6 +31,7 @@ pub enum MTLCPUCacheMode {
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub enum MTLStorageMode {
     Shared  = 0,
+    // macos only
```
kvark: let's put it behind `cfg` as well

src/sampler.rs
```diff
@@ -104,12 +104,14 @@ impl SamplerDescriptorRef {
...
     }
 
+    // requires ios 9.0+ or macos/tvos
     pub fn set_compare_function(&self, func: MTLCompareFunction) {
         unsafe {
             msg_send![self, setCompareFunction:func]
         }
     }
 
+    // TODO: private API
```
kvark: this one is sad as hell

src/device.rs
mtak-: iOS 9.0 has the api `maxThreadsPerThreadgroup`, so this only applies to 8.0-8.4.
I got the numbers from this table:
https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
I don't think I have access to an 8.0 device to test it unfortunately.

src/resource.rs
mtak-: I'll do an audit of the other enums as well.

src/device.rs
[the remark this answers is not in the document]
mtak-: The alternative is to move a lot of this logic into `PrivateCapabilities`. That would be more consistent with the way gfx-backend-metal is already setup.

src/sampler.rs
grovesNL: @kvark did you submit a bug report to clarify this one? I guess it's worth asking about in case it's an oversight

src/device.rs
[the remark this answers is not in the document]
grovesNL: Yeah, I think I'd still prefer to move a lot of the version checks/`respondsToSelector` into `PrivateCapabilities` (not provide default values when a selector is unsupported).

If Apple provides a response to #52 we can think about another approach here, but it would be unfortunate to go through these checks often without the caller being aware.

src/sampler.rs
kvark: I notified Apple in https://github.com/gpuweb/gpuweb/pull/73#issuecomment-416278328 and got no response :/

src/device.rs
```diff
@@ -1538,15 +1601,14 @@ impl DeviceRef {
...
     }
 
-    pub fn new_compute_pipeline_state(&self, descriptor: &ComputePipelineDescriptorRef) -> Result<ComputePipelineState, String> {
-        unsafe {
-            let pipeline_state: *mut MTLComputePipelineState = try_objc!{ err =>
-                msg_send![self, newComputePipelineStateWithDescriptor:descriptor
-                                                               error:&mut err]
-            };
+    #[cfg(feature = "private")]
+    pub unsafe fn new_compute_pipeline_state(&self, descriptor: &ComputePipelineDescriptorRef) -> Result<ComputePipelineState, String> {
```
mtak-: @kvark @grovesNL what do you think about marking private APIs as unsafe?

src/device.rs
```diff
@@ -24,27 +24,63 @@ use std::ptr;
 #[repr(u64)]
 #[derive(Copy, Clone, Debug)]
 pub enum MTLFeatureSet {
+    #[cfg(target_os = "ios")]
```
mtak-: cfging these enums is a little annoying for gfx-backend-metal... but this is more true to the Metal API

examples/caps/main.rs
```diff
@@ -12,11 +12,16 @@ use metal::*;
 fn main() {
     let device = Device::system_default();
 
-    println!("Vendor: {:?}", device.vendor());
-    println!("Family: {:?}", device.family_name());
+    #[cfg(feature = "private")]
```
kvark: it would be easier to just put a single cfg on a block:
```rust
#[cfg(feature = "private")]
{
   ...
}
```

examples/compute/compute-argument-buffer.rs
```diff
@@ -7,7 +7,7 @@
 
 extern crate metal;
 extern crate cocoa;
-#[macro_use] extern crate objc;
+#[macro_use]extern crate objc;
```
kvark: weird change?

src/constants.rs
```diff
@@ -61,77 +68,147 @@ pub enum MTLPixelFormat {
    RGBA32Uint             = 123,
    RGBA32Sint             = 124,
    RGBA32Float            = 125,
+   #[cfg(target_os = "macos")]
```
kvark: uh, this is super wordy and probably inconvenient for the backend implementation
sorry to have you go through this, let's regroup and discuss on gitter a bit more

mtak-:
I toned down the PR a lot. And i think it's ready for a rereview

- Removed `cfg`s
- Kept additional enum variants
- Kept bugfixes
- Kept private feature

src/texture.rs
```diff
@@ -209,6 +209,7 @@ foreign_obj_type! {
 }
 
 impl TextureRef {
+    #[deprecated(since="macOS 10.12, iOS/tvOS 10.0")]
```
kvark: I think the deprecation is meant for Rust APIs and versions, not macOS APIs

mtak-:
Bumped the minor version and changed the deprecation to just `since="0.13.0"`. Thanks for all the feedback!

kvark:
Bors r+

PR #76, opened 2018-12-27
Expose missing MTLDevice::is_removable

repi:
Ran into that this field for detecting eGPUs wasn't exposed, which is a simple add.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1407,6 +1407,16 @@ impl DeviceRef {
         }
     }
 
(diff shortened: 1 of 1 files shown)


kvark:
@repi do you need that patch published, or have more stuff coming first?

repi:
@kvark not at this time, would be great to have a patch published. thanks!

kvark:
https://crates.io/crates/metal/0.13.1

PR #78, opened 2019-01-02
add MTLLanguageVersion 2.1

mtak-:
Missed this from the last PR. `0.14.0` has not been published, so I think this can squeeze in without a version bump.

@kvark Is there anything else in metal-rs that should be updated when new OS Versions come out? If so, I'll try and add it into this PR.

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -111,6 +111,7 @@ pub enum MTLLanguageVersion {
     V1_1 = 0x10001,
     V1_2 = 0x10002,
     V2_0 = 0x20000,
+    V2_1 = 0x20001,
 }
 
 pub enum MTLFunctionConstantValues {}


kvark:
Thanks!
Nothing comes to mind, maybe @grovesNL got something.
Bors r+

grovesNL:
Just the additional enum values, types, selectors, etc. but we don't have to bother block version releases for that IMO. I guess the enum values are the only problematic change because they're technically breaking changes for us.

There are some header diffs online that we can use for reference because the diffs don't seem to be listed in the Metal docs (unfortunately). Although new enum values appear to be listed under "Enumeration Cases" instead of "Constants". See https://developer.apple.com/documentation/metal/mtllanguageversion?language=objc and https://developer.apple.com/documentation/metal/mtltexturetype?language=objc for examples.

PR #79, opened 2019-01-17
Add missing MTLSamplePosition

Michael-Lfx:
cc @kvark

### DIFF
--- a/src/commandbuffer.rs
+++ b/src/commandbuffer.rs
@@ -36,6 +36,13 @@ pub enum MTLCommandBufferError {
     DeviceRemoved = 11,
 }
 
+#[repr(u32)]
+#[allow(non_camel_case_types)]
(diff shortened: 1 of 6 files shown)


src/commandbuffer.rs
```diff
@@ -120,4 +127,10 @@ impl CommandBufferRef {
             msg_send![self, parallelRenderCommandEncoderWithDescriptor:descriptor]
         }
     }
+
+    pub fn compute_command_encoder_with_dispatch_type(&self, ty: MTLDispatchType) -> &ComputeCommandEncoderRef {
```
kvark: Looks like the naming becomes inconsistent with the old `new_compute_command_encoder`.
I think the old one was incorrect and needs fixing, so this isn't a blocker. Filed #80

PR #81, opened 2019-06-12
Fix library name leak

kvark:
cc @scoopr

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ block = "0.1.5"
 foreign-types = "0.3"
 
 [dependencies.objc]
-version = "0.2.3"
+version = "0.2.4"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -29,7 +29,7 @@ block = "0.1.5"
 foreign-types = "0.3"
 
 [dependencies.objc]
-version = "0.2.3"
+version = "0.2.4"
```
grovesNL: Looks like `0.2.6` is available, is there a reason we wanted `0.2.4` specifically?

kvark: no specific reason :) just wanted to get rid of the warnings

grovesNL:
Not sure what happened to bors here
bors r+

PR #84, opened 2019-06-14
Fix Array initialization

kvark:
This PR has essential fixes (2 of them!) to get the argument-buffer example running correctly.


kvark:
Travis is stuck... I went ahead and published https://crates.io/crates/metal/0.15.0

kvark:
Aaand it un-stuck as soon as I complained. Gotta be more patient next time ;)

PR #85, opened 2019-06-14
Fix mem leaks

scoopr:
These commits fix the remaining leaks that I could identify

### DIFF
--- a/src/argument.rs
+++ b/src/argument.rs
@@ -9,7 +9,6 @@ use crate::{Array, MTLTextureType};
 
 use cocoa::foundation::NSUInteger;
 use objc::runtime::{NO, YES};
-use objc_foundation::{INSString, NSString};
 
 #[repr(u64)]
 #[allow(non_camel_case_types)]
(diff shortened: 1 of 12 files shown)


scoopr:
I suppose the objc-foundation removal code is the most invasive, and thinking about it, I guess it could be sidestepped with fixing the various setLabel calls with `&*nslabel` for parameter, and adding a `msg_send![nsname, release]` to `get_function`

kvark:
Could you elaborate on why those dependencies are removed?

src/argument.rs
```diff
@@ -140,8 +139,8 @@ foreign_obj_type! {
 impl StructMemberRef {
     pub fn name(&self) -> &str {
         unsafe {
-            let name: &NSString = msg_send![self, name];
-            name.as_str()
+            let name = msg_send![self, name];
+            crate::nsstring_as_str(name)
```
grovesNL: Is it possible to fix these upstream in objc-foundation instead of removing it?

kvark:
If it's only about `NSString`, it seems like removing 2 external dependencies is a win, even if/when they themselves get the appropriate fixing.
I'm open to bringing them back when times comes, if @grovesNL thinks they are going to be useful.
bors r+

grovesNL:
@kvark that sounds fine to me. objc-foundation also has `NSArray` etc. which would be good to define in a single crate IMO

scoopr:
Yeah, I had removed the objc-foundation wrappers locally before I found a way to workaround the issue, but then felt that in this case less dependencies could be desirable, if it isn't used as much, though I'm ambivalent about it.

PR #87, opened 2019-06-19
Add `newTextureWithDescriptor:iosurface:plane:`

Michael-Lfx:


src/device.rs
```diff
@@ -1380,6 +1380,10 @@ type MTLNewRenderPipelineStateWithReflectionCompletionHandler = extern fn(render
 type MTLNewComputePipelineStateCompletionHandler = extern fn(computePipelineState: id, error: id);
 type MTLNewComputePipelineStateWithReflectionCompletionHandler = extern fn(computePipelineState: id, reflection: id, error: id);*/
 
+#[derive(Debug, Copy, Clone)]
+pub enum _IOSurface { }
```
kvark: We should probably declare it as a full-featured `foreign_obj_type`

src/device.rs
```diff
@@ -1665,6 +1669,14 @@ impl DeviceRef {
     pub fn new_texture(&self, descriptor: &TextureDescriptorRef) -> Texture {
         unsafe { msg_send![self, newTextureWithDescriptor: descriptor] }
     }
+    
+    pub fn new_texture_with_surface_plane(&self, descriptor: &TextureDescriptorRef, iosurface: IOSurfaceRef, plane: NSUInteger) -> Texture {
```
kvark: like with other ref arguments, we should use `&IOSurfaceRef`
I suppose this is directly related to making `IOSurface` to be a real foreign type

src/device.rs
Michael-Lfx: Ok.

src/device.rs
grovesNL: Yeah I mentioned it briefly in #38, should we be using https://github.com/servo/core-foundation-rs/blob/37b82c5564428818726887e7a17450c8bc3227d0/io-surface/src/lib.rs#L19?

xiaopengli89:
How is this PR going? If we want to use `core-foundation`, I can open a new pr.

kvark:
We don't want to use `core-foundation`, only `core-foundation-types`

PR #91, opened 2019-08-07
Add explicit return types for `msg_send!` calls

aleksijuvani:
See gfx-rs/gfx#2939.

### DIFF
--- a/examples/argument-buffer/main.rs
+++ b/examples/argument-buffer/main.rs
@@ -38,6 +38,6 @@ fn main() {
     println!("{:?}", sampler);
 
     unsafe {
-        msg_send![pool, release];
+        let () = msg_send![pool, release];
     }
 }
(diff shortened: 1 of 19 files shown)


src/argument.rs
```diff
@@ -301,22 +301,32 @@ impl ArgumentDescriptor {
 
 impl ArgumentDescriptorRef {
     pub fn set_data_type(&self, ty: MTLDataType) {
-        unsafe { msg_send![self, setDataType: ty] }
+        unsafe {
+            let () = msg_send![self, setDataType: ty];
```
kvark: I don't think these cases need to be fixed? The return type is `()` already, so `msg_send!` result should be inferred

aleksijuvani: Should be taken care of now.

Issue #89, opened 2019-08-04
Handling of `NSArray` for library function names


kasper0406:
In https://github.com/gfx-rs/metal-rs/commit/302d246b6d04741a218a695fc4b86afbbd2401c5 the dependency on `objc-foundation` was removed. Seems like it was mainly used for strings where an ad-hoc helper was used instead. However, at https://github.com/gfx-rs/metal-rs/commit/302d246b6d04741a218a695fc4b86afbbd2401c5#diff-ae230649f72f8f72105a4ebc1edfb748L243 there was a usage of `NSArray` there was translated into an untyped object. 

Before the change the following code worked:
```
    println!("Found the following functions in the Metal library:");
    let function_names = metal_matrix_lib.function_names().to_vec();
    for name in function_names {
        println!("{}", name.as_str());
    }
```
After the change however, the `to_vec` function does not exist.

What are your thoughts on this break? Would you consider re-introducing the dependency to support it, or possibly roll your own array/vec mapping?

I guess this specific call is not that useful in a production setting, but can be quite nifty for debugging during development.

kvark:
Summoning @scoopr for an opinion.

Just to clarify, the breaking itself is fine from the SemVer point of view, since we bumped the version to 0.15. The fact that this code no longer works is a bit sad.

Perhaps, we could expose a method that gets a raw `*const objc::runtime::Object` and gives you `Vec<T>`?

scoopr:
Oh, I don't really have an opinion here. 

I guess I missed between the initial solution searching and the final PR forming that I had actually changed a public method signature. I suppose a helper method of sorts would make sense here.

Alternatively, a return type of `Vec<String>` would probably be ideal from the api use side, but of course wouldn't be quite zero cost. Also I suppose a `impl Iterator` would work too

PR #92, opened 2019-08-16
Fix memory leak in `nsstring_from_str`

aleksijuvani:


src/lib.rs
```diff
@@ -47,10 +47,8 @@ fn nsstring_from_str(string: &str) -> *mut objc::runtime::Object {
     let cls = class!(NSString);
     let bytes = string.as_ptr() as *const c_void;
     unsafe {
-        let obj: *mut objc::runtime::Object = msg_send![cls, alloc];
-        let obj: *mut objc::runtime::Object = msg_send![obj, initWithBytes:bytes
-                                                    length:string.len()
-                                                    encoding:UTF8_ENCODING];
+        let mut obj: *mut objc::runtime::Object = msg_send![cls, alloc];
+        obj = msg_send![obj, initWithBytes:bytes length:string.len() encoding:UTF8_ENCODING];
```
kvark: how is this fixing anything? your change looks equivalent to the old code

aleksijuvani: My bad! I was seeing a memory leak show up in Instruments pointing to this function, and it disappeared upon making this change. However, upon further inspection, it looks like this was actually already fixed by #85, which hasn't been published yet. I had assumed that gfx-backend-metal was using the latest commit. I thought I had tested this before making the change, but I must've had the dependency pointing at the wrong version at the time.

aleksijuvani: I would suggest publishing a new version of this crate so that we can take advantage of the latest changes in gfx-backend-metal.

kvark: I'd like to release, but it looks like #89 needs to be resolved first in some way

PR #102, opened 2019-10-29
Remove objc_exception dependency

kvark:


kvark:
Closing as unsafe, as indicated by @grovesNL 
Opened https://github.com/SSheldon/rust-objc-exception/issues/4 instead

PR #112, opened 2020-02-13
added build script

adamnemecek:


build.rs
```diff
@@ -0,0 +1,32 @@
+use std::process::Command;
+
+fn build_example_shaders(example: &str) {
+    let example = example.to_string();
+    let rerun = format!("cargo:rerun-if-changed=examples/{}/shaders.metal", example);
+    println!("{}", rerun);
```
kvark: why not just put the stuff into `println!`?

build.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    let shader = format!("examples/{}/shaders.metal", example);
+    let air = format!("examples/{}/shaders.air", example);
+    let lib = format!("examples/{}/shaders.metallib", example);
+
+    let air_status = Command::new("xcrun")
```
kvark: I am sort of worried about this. Would any dependencies need to be able to run `xcrun` now during the build process?

Imagine, say that gfx-rs is built as a dependency, cross-compiled from linux to macos. I don't think we even have `xcrun` in this case. So this approach isn't compatible with cross-compiling at all?

See also - https://github.com/gfx-rs/gfx/commit/cc0930b49e778f04ad523abe5a20736d34dc4fea#diff-a0fb0c848201db4ba17b14b5299fd08a

Perhaps the solution here is to only do this when building in some specific conditions, e.g. when `#[test]` is set.

adamnemecek: Yes, they would have to be able to run xcrun. Do you want to put the binaries back into version control?

kvark: I'm thinking if the best way would be to just do this inside the example code (calling `xcrun`).

adamnemecek: How about making a separate binary in /bin?

build.rs
```diff
@@ -0,0 +1,36 @@
...
+        .args(&["-sdk", "macosx", "metal", "-c", &shader, "-o", &air])
+        .status()
+        .expect("failed to execute metal compiler");
+    if !air_status.success() {
+        panic!("Failed to compile .metal -> .air");
+    }
+
+    let metal_status = Command::new("xcrun")
```
simlay: I'm not sure it was supported before but I'm guessing this will break iOS support? If you do this pattern you'll preserve iOS support: https://github.com/flutter-rs/flutter-rs/blob/136bb162a895630d2b75dc674a60da51a6e467d8/flutter-engine-sys/build.rs#L48

PR #113, opened 2020-02-20
updated winit to 0.21

adamnemecek:


kvark:
@adamnemecek in terms of workflow, we typically create branches for work. You'd not want to do a feature in `master`, since it's inconvenient to switch between features or go back to a working code.

kvark:
Could you squash please? Also, I assume you addressed the compile warnings.

kvark:
Thanks!
bors r+

PR #116, opened 2020-02-22
indexing (not working currently)

adamnemecek:


examples/window/main.rs
```diff
@@ -34,20 +34,26 @@ fn prepare_pipeline_state<'a>(device: &DeviceRef, library: &LibraryRef) -> Rende
     pipeline_state_descriptor.set_vertex_function(Some(&vert));
     pipeline_state_descriptor.set_fragment_function(Some(&frag));
     pipeline_state_descriptor
-        .color_attachments()
-        .object_at(0)
-        .unwrap()
+        .color_attachments()[0]
+        // .object_at(0)
```
kvark: let's clean up all these comments

examples/window/main.rs
```diff
@@ -34,20 +34,26 @@ fn prepare_pipeline_state<'a>(device: &DeviceRef, library: &LibraryRef) -> Rende
...
         .set_pixel_format(MTLPixelFormat::BGRA8Unorm);
 
     device
         .new_render_pipeline_state(&pipeline_state_descriptor)
         .unwrap()
 }
 
+fn ppd(desc: &mut RenderPassDescriptorRef) {
+    // let color_attachment = &desc.color_attachments()[0];
+    desc.color_attachments()[0] = desc.color_attachments()[0].to_owned();
```
kvark: We can live with the fact this doesn't work. It doesn't have to be provided - users can still do `set_object_at`, right?

src/pipeline/compute.rs
```diff
@@ -261,13 +261,28 @@ foreign_obj_type! {
...
+    type Output = AttributeDescriptor;
+
+    fn index(&self, index: NSUInteger) -> &Self::Output {
         unsafe { msg_send![self, objectAtIndexedSubscript: index] }
     }
+}
 
-    pub fn set_object_at(&self, index: usize, buffer_desc: Option<&AttributeDescriptorRef>) {
-        unsafe { msg_send![self, setObject:buffer_desc atIndexedSubscript:index] }
+impl std::ops::IndexMut<NSUInteger> for AttributeDescriptorArrayRef {
```
kvark: It seems to me that this isn't actually needed. Today, metal-rs doesn't enforce thread guarantees (unfortunately!) so all the methods are `&self`. Therefore, there isn't a case where we'd need `&mut self`.

In the future, it will definitely be desired to make us properly Send/Sync/Clone/&mut, but this is out of scope of this PR.

kvark:
@adamnemecek ping. Are you interested to land this one day?

adamnemecek:
I’ll revisit this.

PR #120, opened 2020-03-01
version = 0.17.3, added default for types.rs

adamnemecek:


kvark:
I'm afraid #115 was a breaking change, and we can't just bump the patch version. Let's bump the minor?

PR #127, opened 2020-03-23
Dependency version updates

kvark:


grovesNL:
Oops, I assume we want to do the version bump in here
bors r-

grovesNL:
Actually I just noticed that we've already bumped the version to 0.18 :)
bors r+

PR #129, opened 2020-04-03
Add headless rendering example

chinedufn:
Hey!

This PR adds an example of how to headlessly render and save to an image.

I wanted to add this example while it was fresh on my mind because it took me a few hours to figure out how to do it.

This is my first time writing Metal(can't believe I can have my first metal experience with Rust!!) so feel free to point out anything silly that I might have done.

I tried to stay close to the existing samples so hopefully nothing looks too off.

Happy to make any changes that you request - cheers!


examples/headless-render/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    img.save(&out_file).unwrap();
+
+    unsafe {
+        libc::free(img_bytes as *mut libc::c_void);
+    };
+
+    println!("Image saved to {:?}", out_file);
+}
+
+fn create_layer(device: &Device) -> CoreAnimationLayer {
```
kvark, line 143: Why would you want to create a CAL if you are featuring a headless mode? Wouldn't you just want to create a regular Metal texture, render into it, and read it back instead?

examples/headless-render/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    device
+        .new_render_pipeline_state(&pipeline_state_descriptor)
+        .unwrap()
+}
+
+fn create_vertex_buffer(device: &DeviceRef) -> Buffer {
+    device.new_buffer_with_data(
+        VERTEX_ATTRIBS.as_ptr() as *const _,
+        (VERTEX_ATTRIBS.len() * mem::size_of::<f32>()) as u64,
+        MTLResourceOptions::CPUCacheModeDefaultCache | MTLResourceOptions::StorageModeManaged,
```
kvark, line 178: this should be Private

examples/headless-render/main.rs
```diff
@@ -0,0 +1,189 @@
...
+        slice.copy_from_slice(&img_slice[idx * 4..idx * 4 + 4]);
+        let slice = [slice[2], slice[1], slice[0], slice[3]];
+
+        img.put_pixel(x as u32, y as u32, Rgba(slice));
+    }
+
+    let out_file =
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("examples/headless-render/out.png");
+
+    img.save(&out_file).unwrap();
```
kvark, line 134: could we hook up something lighter than `image` crate? i.e. just the `png` crate?

examples/headless-render/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    let img_slice: &[u8] = unsafe { std::slice::from_raw_parts(img_slice, TOTAL_BYTES as usize) };
+
+    let mut img = DynamicImage::new_rgba8(VIEW_WIDTH as u32, VIEW_HEIGHT as u32);
+
+    for idx in 0..(TOTAL_BYTES / 4) {
+        let x = idx % VIEW_WIDTH as usize;
+        let y = idx / VIEW_HEIGHT as usize;
+
+        let mut slice = [0; 4];
+        slice.copy_from_slice(&img_slice[idx * 4..idx * 4 + 4]);
```
kvark, line 125: This looks like the second copy we do. We should only have one copy of the data.

examples/headless-render/main.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    command_buffer.commit();
+
+    command_buffer.wait_until_completed();
+
+    save_image(&texture);
+}
+
+fn save_image(texture: &TextureRef) {
+    let img_bytes = unsafe {
```
kvark, line 85: this should probably be allocated as rust `Vec` right away

examples/headless-render/main.rs
chinedufn: Removed thanks!

examples/headless-render/main.rs
chinedufn: Are we allowed to read the bytes back on the CPU in private mode?

I changed this to `MTLResourceOptions::CPUCacheModeDefaultCache | MTLResourceOptions::StorageModePrivate` and got a seg fault.

Mind giving me some guidance here?

examples/headless-render/main.rs
chinedufn: Yo!!! I've been using `image` for two years for only PNGs and it felt super heavy for my needs.

I don't know why I never googled "rust png" ...

Love this - thanks - changed.

examples/headless-render/main.rs
chinedufn: Removed the extra copy

examples/headless-render/main.rs
chinedufn: I think I removed the need to allocate at all in https://github.com/gfx-rs/metal-rs/pull/129/commits/478ff7c34702f119d1c8a8e1afebc572ed883fd3 , but let me know if I'm missing something

chinedufn: Oh thanks - I'm learning some good stuff in this PR -> 16a7aac

chinedufn:
@kvark alright I'm addressed all of your concerns so far. Thanks for the review

examples/headless-render/main.rs
kvark: this is a vertex buffer, why do you need to read it back?

examples/headless-render/main.rs
```diff
@@ -0,0 +1,168 @@
...
+                width: VIEW_WIDTH,
+                height: VIEW_HEIGHT,
+                depth: 1,
+            },
+        },
+        0,
+        VIEW_WIDTH * 4,
+    );
+
+    let mut image = unsafe { Vec::from_raw_parts(img_bytes as *mut u8, TOTAL_BYTES, TOTAL_BYTES) };
```
kvark, line 102: we shouldn't cast it to a vec here, it's unsafe and could very well be unsound :)
instead, just allocate it as `vec![size]`, then you can get a pointer to it for `get_bytes`.

examples/headless-render/main.rs
kvark: it's used by `image`, so it's not really an alternative, just a lightweight way to access the same thing :)

examples/headless-render/main.rs
chinedufn: Ah for some reason I was thinking it was a texture - whoops

At any rate, it seg faults when I set it to private

chinedufn: Any suggestions on how to best debug that?

kvark: ok, perhaps it doesn't like private data initialized on CPU side, nevermind then!
let's keep the Managed there

examples/headless-render/main.rs
kvark: this still needs to just do `vec![TOTAL_BYTES]` instead of going through `libc::malloc`

examples/headless-render/main.rs
chinedufn: Ahh didn't realize you could cast between the pointers - alright boom I'm getting the hang of it. Fixed!

chinedufn:
@kvark all feedback so far addressed

examples/headless-render/main.rs
```diff
@@ -0,0 +1,160 @@
...
+                depth: 1,
+            },
+        },
+        0,
+        VIEW_WIDTH * 4,
+    );
+
+    for idx in 0..(TOTAL_BYTES / 4) {
+        let idx = 4 * idx;
+        image.swap(idx, idx + 2);
```
kvark, line 98: can we just use `RGBA8Unorm` format instead and avoid this whole swizzle loop?

chinedufn: Ah I tried using RGBA when I was playing with the window example and ran into some issues - forgot to try it for this example

Worked - thanks! f36dade

chinedufn:
SQUASHED

chinedufn:
Alright so I looked into the seg fault when trying to use the private storage mode in this example.

It was because private is meant for resources that are *never* accessed by the CPU, not even to populate them with initial data. 

So it's mostly meant for resources that are populated as intermediary data during render passes.

image

via -> https://developer.apple.com/documentation/metal/setting_resource_storage_modes/choosing_a_resource_storage_mode_in_macos

aleksijuvani:
yes, private storage mode is the same thing as device local memory in vulkan, meaning that you gotta use a staging buffer to copy data into it on the gpu. we can't initialize these with `newBufferWithBytes:length:options:`

kvark:
I wasn't sure if Metal run time does that for us. It could easily allocate a temporary staging buffer and issue a copy, while allowing the storage mode for the target to be Private.

PR #130, opened 2020-04-04
send + sync

adamnemecek:


src/buffer.rs
```diff
@@ -59,4 +62,4 @@ impl BufferRef {
             msg_send![self, addDebugMarker:name range:range]
         }
     }
-}
+}
```
kvark, line 65: that should be reverted :)

adamnemecek: I need it for textures at least.

kvark:
> I need it for textures at least.

I understand you may want it for some things. So do us in gfx-rs! But advertising Send+Sync at the metal-rs level just seems wrong, since we know it's false.

What you can easily do instead on your side is wrap the types, i.e.
```rust
struct MyTexture(metal::Texture);
unsafe impl Send for MyTexture;
unsafe impl Sync for MyTexture;
```

It gives you all the same abilities, but you are opting into the unsafety explicitly.

PR #136, opened 2020-04-11
Add padding to headless render example

chinedufn:
It looks like both the window example and headless render example aren't rendering properly.

The vertex data suggests that you should see a triangle with one red vertex, one blue and one green. 

That's not what you see when you run the examples though.

It seems like the interleaved position/color attributes are each expected to be padded to 16 bytes. Here's a before and after, after is when I added the padding that you can see in this PR's first commit.

## Before:

image

## After:

image

---

Doing some googling to see if there's a way to interleave the attributes without needing to pad the data. 

This article further suggests that the data needs to be padded

image

via: https://metalbyexample.com/vertex-descriptors/

---

Not sure if there's a better approach that would allow us to remove all padding. Just let me know!


chinedufn:
If this approach is approved (or if we nail down a better one) I can apply it to the window example before landing this PR

chinedufn:
Found it 
image

via: https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf Section 2.2.3

---

I've updated the PR to use the packed versions of the attributes and updated the window example to use an equilateral triangle.

examples/window/compile-shader.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+xcrun -sdk macosx metal -c shaders.metal -o shaders.air
```
kvark, line 5: please add this to a makefile at the root instead of a separate shell script

chinedufn: Turns out it was already in the Makefile, done 87a8ce3

PR #138, opened 2020-04-17
Support Vertex Amplification.

abd45:
This PR adds support for Vertex Amplification which is a new feature in Metal. Details of vertex amplification is available here: https://developer.apple.com/documentation/metal/generating_multiple_output_vertex_streams_from_one_input_stream?language=objc

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -113,6 +113,13 @@ pub struct MTLDrawIndexedPrimitivesIndirectArguments {
     pub baseInstance: u32,
 }
 
+#[repr(C)]
+#[derive(Copy, Clone, Debug)]
+pub struct VertexAmplificationViewMapping {
+    pub renderTargetArrayIndexOffset: u32,
+    pub viewportArrayIndexOffset: u32,
+}
+
 pub enum MTLCommandEncoder {}
 
 foreign_obj_type! {
@@ -137,9 +144,7 @@ impl CommandEncoderRef {
     }
 
     pub fn end_encoding(&self) {
-        unsafe {
-            msg_send![self, endEncoding]
-        }
+        unsafe { msg_send![self, endEncoding] }
     }
 
     pub fn insert_debug_signpost(&self, name: &str) {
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -255,6 +258,16 @@ impl RenderCommandEncoderRef {
         }
     }
 
+    pub fn set_vertex_amplification(
```
kvark, line 261: should probably be called `set_vertex_amplification_count` for consistency

src/encoder.rs
```diff
@@ -255,6 +258,16 @@ impl RenderCommandEncoderRef {
...
+        &self,
+        count: NSUInteger,
+        view_mappings: &[VertexAmplificationViewMapping],
```
kvark, line 264: should be optional?

abd45: Do you mean to add `#[cfg(feature = "private")]` here ?

kvark: no, I mean `Option<&[...]>`

abd45: Hmm.. The user of this method has to set the structure array and then provide the structure array as parameter to the set_vertex-amplification_count function. If it's an Option, it means we are allowing None for view mapping structure without which the rendering will be same as original. Should we allow None for [VertexAmplificationViewMapping] when an user already chose to use this function? What Say?

kvark: We are providing the bindings to Metal API, so in case of doubt just follow the upstream:
> viewMappings
>    An optional parameter that specifies per-output offsets into the layer and viewport lists. If you provide a value for this parameter, you must provide the same number of entries as the count you specified.

abd45: Thanks @kvark for the clarification. I updated the PR with changes.

src/encoder.rs
```diff
@@ -255,6 +258,16 @@ impl RenderCommandEncoderRef {
         }
     }
 
+    pub fn set_vertex_amplification_count(
+        &self,
+        count: NSUInteger,
+        view_mappings: Option<&[VertexAmplificationViewMapping]>,
+    ) {
+        unsafe {
+            msg_send! [self, setVertexAmplificationCount: count viewMappings: view_mappings.unwrap().as_ptr()]
```
kvark, line 267: this can't be `unwrap()` :) This would be `map_or(ptr::null(), |vm| vm.as_ptr())`

abd45: Aha! My bad! Did not consider the case when the mapping is None! Thank you. Good learning.

kvark:
Bors r+

PR #146, opened 2020-05-04
Use more persistent shader type cache dir

chinedufn:
Because the `OUT_DIR` environment variable can change, the shader bindings cache would get invalidated sometimes leading to a recompilation of shader bindings.

Generating the bindings is a slow process and should be avoided.

This PR creates a directory in the crate to use for caching since having a consistent cache directory means shader type re-compilation becomes rare.

To further reduce the cost of these bindings, I changed the `shader_types.rs` file to only include `simd/vector_types.h` instead of all of `simd/simd.h`. 

This takes a `cargo build -p texture` after editing `shader_types.h` from 16seconds to around 9.7 seconds on my machine (effectively all of that time is spent generating the Rust bindings).

We're really only using one type from that file (vector_float2), but I didn't spend time further investigating how to reduce the amount of simd that we're needlessly compiling (I did investigate rust bindgen whitelists/blacklists but those just control what ends up in the final generating bindings. So no time is saved by white/blacklisting things as far as I can tell).

### DIFF
--- a/examples/texture/.gitignore
+++ b/examples/texture/.gitignore
@@ -1,1 +1,2 @@
 shaders.metallib
+shader_types_hash
--- a/examples/texture/build.rs
+++ b/examples/texture/build.rs
@@ -106,5 +106,5 @@ fn save_shader_types_hash(hash: u64) {
 }
 
 fn shader_types_hash_file() -> PathBuf {
-    PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("shader_types_hash")
+    PathBuf::from(std::env::var("CARGO_MANIFEST_DIR").unwrap()).join("shader_types_hash")
 }
--- a/examples/texture/shader_types/shader_types.h
+++ b/examples/texture/shader_types/shader_types.h
@@ -5,7 +5,10 @@
 #ifndef shader_types_h
 #define shader_types_h
 
-#include <simd/simd.h>
+// Used to get the definition of vector_float2
+// If you need to import more types, simd is typically located at:
+// /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/simd
+#include <simd/vector_types.h>
 
 typedef enum VertexInputIndex {
     VertexInputIndexVertices = 0,


kvark:
> Because the OUT_DIR environment variable can change, the shader bindings cache would get invalidated sometimes leading to a recompilation of shader bindings.

I'm curious about more details of this scenario. When would `OUT_DIR` change, and what would you be building in this case that would require recompilation of the shaders?

My limited understanding is that the build script is for the examples only, it's not involved in any dependencies of the library, and thus the only scenario it's used in is when somebody checks out the repo and tries out the examples. Moreover, the extra time it would take to build if OUT_DIR was changed wouldn't be a problem, because we are talking about a developer who is just trying to run the examples.

chinedufn:
It looks like `OUT_DIR` [can change when certain metadata about the package changes.](https://github.com/rust-lang/cargo/blob/6a2d62fb402d1cd47d8eadf890522fdba3044b48/src/cargo/core/compiler/context/compilation_files.rs#L239-L247).

(When finding that link I didn't look into what all the metadata that is considered is)

I also don't know whether or not there are ever times where `OUT_DIR` gets cleared. 

But I do know that even as I was working on this PR I was getting recompiles of the shader types occasionally when I most definitely did not touch them.

What prompted this PR was I came back to the `texture` `build.rs` to adapt it to a different project that I'm working on and when I built the project the build script ran even though I hadn't changed anything.

> Moreover, the extra time it would take to build if OUT_DIR was changed wouldn't be a problem, because we are talking about a developer who is just trying to run the examples.

In the lens of the example, yeah - which is why I almost didn't even include the caching logic in the original PR.

But then I remembered that, naturally, most developers that are getting started with metal and Rust will look at existing examples and copy from them (the entire point of an example, really).

So if one of the places they'll look first (metal-rs) has an example that encourages the wrong idea that becomes something that starts to trail through other codebases over time.

So yeah if it was just about this example the caching isn't at all useful. I'm just thinking that this example will be copied by >0 people so we should give them the best that we reasonably can.

---

In short - to me we're trading a small cost (adding a single line to `.gitignore`), for a (relatively.. emphasis on relatively) large win (bringing the number of times that a build *unnecessarily* takes an extra 10+ seconds down to 0)

---

So yeah - that's my thinking. Sorry for not including more of it in the original PR description.

As usual let me know if you think I'm off base and you see it completely differently!

kvark:
> But I do know that even as I was working on this PR I was getting recompiles of the shader types occasionally when I most definitely did not touch them.

If this is unrelated to the `OUT_DIR`, we'd still need to investigate.

kvark:
bors retry

chinedufn:
> If this is unrelated to the OUT_DIR, we'd still need to investigate.

Ah sorry I meant that I was getting that before I stopped using OUT_DIR

This recompilation no longer occurs.

kvark:
Perhaps you could rebase this to kick off new GH actions round?

chinedufn:
Done

chinedufn:
Looks like bors never merged this

kvark:
bors retry

kvark:
actually, CI was all green, IIRC

PR #151, opened 2020-07-05
Add guide to debugging in Xcode

### LINKED ISSUE #148: Add documentation describing how to set up Xcode debugging
I’d like to add an example of taking a Metal app that’s completely written in Rust and debugging it in Xcode.

I’ve never actually done this - but I have some thoughts on how to approach it.

_On mobile so might have typos_

1. Add an example that just imports one of the other examples (such as the textured quad example). So the imported example would have to expose a “run” method. Use an environment variable to make the shader compilation include debug info. (So that when people copy paste the build script for their real project they aren’t accidentally generating debug builds without realizing)

2. Our Xcode debugging example would compile to a dylib that exposed a single method to run it

3. Xcode example would have the smallest possible Xcode project that links to and runs the Rust generated dylib

4. In the README of the example we include a screenshot or two of an Xcode debug session as well as any important notes / caveats if there end up being any

The nice part would be that the Xcode project would be immediately re-usable for any application since it does nothing but run a dylib.

One reason that I can think of against adding this example is that in general Xcode tends to feel bloaty (or at least that was my experience 6 or so years ago). I think that this is mitigated by having a project that does nothing more than run a linked library.

Another question might be - does this belong in metal-rs or a separate repo?

My thinking would be that being able to debug your shaders can be such a timesaver as well as such a cool demo as well as something you might not have thought of (I can still use Xcode tools with my entirely Rust app?!?!) that it belongs alongside the core examples.

Once people know how to run Xcode tools on their Rust app this means they now also know how to profile shaders, as well as anything else you might want to lean on Xcode’s user interface for (anything you want to visualize).

So this one example would unlock all Xcode tooling for the developer.

To repeat - I’ve never actually done this so it sounds good on paper but who knows what gotchas might exist. 

But that’s the point of the example. We discover the solutions to those and present it in a nice tidy little example.

Let me know your thoughts and where you agree or disagree!

chinedufn:
Closes #148


guide/debugging-in-xcode/README.md
```diff
@@ -0,0 +1,37 @@
+# Debugging in Xcode
```
chinedufn, line 1: [rendered](https://github.com/gfx-rs/metal-rs/blob/fcd87983cf5171695fba04053b7e3d7c67c516d1/guide/debugging-in-xcode/README.md)

chinedufn:
It turns out that you can also capture GPU frame programatically (without using Xcode) and write them to a file.

Then later just open that file and it will launch in Xcode.

https://developer.apple.com/documentation/metal/frame_capture_debugging_tools/capturing_gpu_command_data_programmatically

I left a TODO to add this to the debugging docs

adamnemecek:
This is very timely. Are you on discord or twitter or something? I think that we are like the only two rust developers who use metal directly, we should form some sort of union.

chinedufn:
Haha I'm not on either but my email is in my GitHub profile if you ever want to chat Metal and all that good stuff

chinedufn:
Ha so if it was *only* up to me I'd prefer an mdbook. I went ahead and assumed that you might feel that was too heavy handed. So I threw in the markdown file.

I like being able to #include the real example code in the guide a la -> https://rustwasm.github.io/docs/wasm-bindgen/examples/web-audio.html .

But that aside. As someone who primarily works offline - having guides be in tree is helpful. I like to be able to just clone a repo and have everything that I need.

---

Now, that all being said. I have no issues with going with the wiki!

My only goal/vision is that other people get to experience the joy of targeting Metal with Rust. It's a super smooth experience and I want others to know it!

I'd like a future where someone building a production-grade Metal application is seriously considering Rust + `metal-rs`.

To me a major gap for that future is for someone to land in `metal-rs` and immediately know how to get running. I want us to hold people's hands. "Here's how the bindings work (lightly explain `msg_send!`, etc)." "Here's some assurance that you aren't missing anything that you would have targetting Metal from Swift." etc.

Ok so slightly off-topic rant over.

I say all of that to say.

I'm cool with a wiki albeit not my first choice. Could you create a wiki for the repo and I'll get this moved?

Cheers!

kvark:
Ok, that's fine by me :)
If you don't have write access to the wiki, please let me know.

chinedufn:
Whenever I try to go to the wiki it redirects me to the repo homepage

kvark:
@chinedufn please try again now!

PR #154, opened 2020-07-05
removed duplicate definition of set_label on Buffer since it's alread…

adamnemecek:
…y defined on Resource, added missing methods to Resource

### DIFF
--- a/src/resource.rs
+++ b/src/resource.rs
@@ -108,7 +108,8 @@ impl ResourceRef {
         unsafe { msg_send![self, setPurgeableState: state] }
     }
 
-    pub fn allocated_size(&self) -> u64 {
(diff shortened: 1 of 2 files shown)


src/resource.rs
```diff
@@ -102,4 +107,8 @@ impl ResourceRef {
     pub fn set_purgeable_state(&self, state: MTLPurgeableState) -> MTLPurgeableState {
         unsafe { msg_send![self, setPurgeableState: state] }
     }
+
+    pub fn allocated_size(&self) -> u64 {
```
kvark, line 111: Let's use the actual `NSUInteger` type
also, would be good to have a comment saying this is available on macOS 10.13+ and iOS 10.11+

PR #157, opened 2020-07-10
Remove core-graphics dependency

### LINKED ISSUE #156: Remove core-graphics dependency
Right now `core-graphics` is used on these two lines (excluding in examples. I'm talking strictly `metal-rs` source)

https://github.com/gfx-rs/metal-rs/blob/c17acf1aa38e05cf5fe293f6fc91e682ee71b2fb/src/lib.rs#L27-L28

Effectively we're using it to define type aliases for a couple of scalar types.

Given that the chances of these aliases ever changing is roughly 0%, I'm proposing that we ditch the `core-graphics` dependency and instead just copy-paste those two types inline in `metal-rs`.

The end result would be inlining the following code:

```rust
#[cfg(target_pointer_width = "64")]
type CGFloat = f64;
#[cfg(not(target_pointer_width = "64"))]
type CGFloat = f32;

#[repr(C)]
#[derive(Clone, Copy, Debug, Default, PartialEq)]
struct CGSize {
    pub width: CGFloat,
    pub height: CGFloat,
}
```

Am I missing anything here?

chinedufn:
Closes #156

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,10 +24,26 @@ use std::ops::Deref;
 use std::os::raw::c_void;
 
 use cocoa::foundation::NSUInteger;
-use core_graphics::base::CGFloat;
-use core_graphics::geometry::CGSize;
 use foreign_types::ForeignType;
 use objc::runtime::{Object, BOOL, NO, YES};
+
+#[cfg(target_pointer_width = "64")]
+pub type CGFloat = f64;
+#[cfg(not(target_pointer_width = "64"))]
+pub type CGFloat = f32;
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct CGSize {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -24,11 +24,27 @@ use std::ops::Deref;
 use std::os::raw::c_void;
 
 use cocoa::foundation::NSUInteger;
-use core_graphics::base::CGFloat;
-use core_graphics::geometry::CGSize;
 use foreign_types::ForeignType;
 use objc::runtime::{Object, BOOL, NO, YES};
 
+#[cfg(target_pointer_width = "64")]
+type CGFloat = f64;
```
kvark, line 31: let's make this public too, otherwise using `CGSize` becomes complicated from the outside

chinedufn: Done!

PR #160, opened 2020-07-17
[0.18] Bump dependencies

kvark:
This is a little bit risky, since there are parts of `cocoa` that we expose in public API (e.g. `NSUInteger`), so updating the breaking version of it is technically also a breaking change... But it's fine as long as both versions of cocoa just typedef it to the same thing, e.g. `u64`.
If there is a way in Rust to know *all* of the public uses of a dependency, it would be handy here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "metal"
-version = "0.18.0"
+version = "0.18.1"
 description = "Rust bindings for Metal"
 documentation = "https://docs.rs/crate/metal"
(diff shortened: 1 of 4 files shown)


kvark:
@chinedufn I yanked v0.18.1 and reverted the `cocoa` update - it caused linking errors with `NSRange` type :(
We'll just need to release a new breaking version.

chinedufn:
Yikes - thanks for reverting.

Cool, `0.19.0` makes sense to me

PR #164, opened 2020-07-29
added method MTLBlitCommandEncoder::generate_mipmaps

adamnemecek:
https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400748-generatemipmaps

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -706,6 +706,10 @@ impl BlitCommandEncoderRef {
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -706,6 +706,10 @@ impl BlitCommandEncoderRef {
         }
     }
 
+    pub fn generate_mipmaps(&self, texture: &TextureRef) {
+        unsafe { msg_send![self, for: texture] }
```
kvark, line 710: message should be `generateMipmapsForTexture`

Issue #144, opened 2020-05-02
Make function parameter names and order consistent with Apple's documentation


chinedufn:
In Apple's documentation, the order of function parameters might look like this:

```
func replace(region: MTLRegion, 
 mipmapLevel level: Int, 
   withBytes pixelBytes: UnsafeRawPointer, 
 bytesPerRow: Int)
```

whereas in `metal-rs` it might look like this

```rust
    pub fn replace_region(
        &self,
        region: MTLRegion,
        mipmap_level: NSUInteger,
        stride: NSUInteger,
        bytes: *const std::ffi::c_void,
    ) {
        // ...
    }
```

I found that the order differences (last two parameters swapped) and name differences (`stride` vs. `bytes_per_row`) made it difficult to cross reference Apple's Metal docs when working with `metal-rs`.

My understanding is that these were all written by hand over time (I'm just basing this assumption on the first few commits from 2016 https://github.com/gfx-rs/metal-rs/commits/master?after=09433e64682ffe4498988118c062e608a4971eb6+220)

Given that - it's probably quite a bit of manual work and breaking changes to get consistent (or at least to the degree possible) with the official metal APIs.

How does `metal-rs` typically think about breaking changes? Is this a no-go? Or could this fit in at some point in the future?

I could use some insight here.

Cheers!

kvark:
We are totally fine with breaking changes like that, but we adhere to semver like any other crate. Therefore, breaking changes will go into the breaking revision.

I agree in this case that we should rename the arg to `bytes_per_row`. There are other cases, which can be more complex, i.e. `setBuffer` in Metal accepts the slot in a non-obvious position. We can discuss all of them here.

chinedufn:
Sounds good thanks for explaining.

Sweet. I'll bring them up here on a case by case basis whenever I'm working with `metal-rs` and run into one.

kvark:
I aligned the texture API in 461148ed39a14e319992a2c1478d9a978cc32ffd

PR #165, opened 2020-07-30
added constructors to MTLRegion

adamnemecek:


src/types.rs
```diff
@@ -31,6 +31,37 @@ pub struct MTLRegion {
     pub size: MTLSize,
 }
 
+impl MTLRegion {
+    #[inline]
```
chinedufn, line 35: I'm curious to hear more about where you feel the line should be for abstractions around the raw metal APIs?

Mainly looking to better understand the motivation / thought process here.

More specifically, why should this land?

Cheers!!

chinedufn: Ah I see https://developer.apple.com/documentation/metal/1516011-mtlregionmake1d

Mind making names consistent with Metal and linking to the Metal docs as a doc comment?

adamnemecek: As far as naming, you need to strike a balance between the metal idioms and rust idioms. What do you suggest these be called? We already diverge from the make prefix e.g. new_render_command_encoder is called makeRenderCommandEncoder in metal.

chinedufn: I think one of the most important flows to support is

1. Person looks up Apple’s Metal Docs for how to do something

2. Person looks for the equivalent in metal-rs

3. Person finds what they need quickly

Given that goal (I’m just speaking for myself, you might think that’s a silly goal!) I’d say that the only changes we should make to function names or parameters is 

1. snake casing because it’s more rusty and you’ll still be able to find what you’re looking for

2. More appropriate types such as accepting a &Path instead of an NSString when we need a file path.

As such, I’d personally lean towards calling these make_1d, make_2d and make_3d.

I’d say that the divergence in the render command encoder creator fits under the umbrella of https://github.com/gfx-rs/metal-rs/issues/144

These are just my thoughts, will wait on @kvark

adamnemecek: So these should be functions or constructors?

chinedufn: Ah sorry, I meant MTLRegion::make_1d etc constructors

adamnemecek: I would still lean towards the new nomenclature since there isn't a single instance in metal-rs of a function starting with "make".

adamnemecek: @kvark suggests renaming all functions to "make" (pun intended) to be in line with metal. How about we merge this in and rename it with the rest if we decide to do that?

chinedufn: I just appreciate this PR in the first place so - I won’t stand in the way! That plan sounds fine to me.

Maybe just still add the doc comment linking to the corresponding Metal docs?

I think we should do that for all constructors/functions/methods/types/etc going forwards.

chinedufn: Oh and I still think these should at the very least be 1d 2d instead of d1 d2 ... but again ... I won’t stand in the way and bike shed this !

adamnemecek:
Yeah I was actually just trying to rename that.

PR #166, opened 2020-07-31
mtlregion constructors

adamnemecek:
This was originally #165 but I pushed more than I wanted so here we go again.


kvark:
Merging with a note that we want to revisit the naming for consistency with Metal.

PR #167, opened 2020-07-31
examples/window draws a clear, instanced rect

adamnemecek:
Shows off some APIs that other examples don't, such as `draw_primitives_instanced` and `set_scissor_rect`. Also, I never remember the order in which `rect_vert` should return the single vertices so here it's preserved for the future.


examples/window/main.rs
```diff
@@ -19,9 +19,36 @@ use winit::{
     event_loop::ControlFlow,
 };
 
-fn prepare_pipeline_state<'a>(device: &DeviceRef, library: &LibraryRef) -> RenderPipelineState {
-    let vert = library.get_function("triangle_vertex", None).unwrap();
-    let frag = library.get_function("triangle_fragment", None).unwrap();
+#[repr(C)]
+struct Rect {
+    pub x: f32,
```
kvark, line 24: what's the point of all the `pub` in here?

adamnemecek: habit lol.

PR #168, opened 2020-08-01
added RenderCommandEncoder set_vertex_value + set_fragment_value

adamnemecek:
I realize this is probably a matter of taste but the original API is really unwieldy in Rust. @chinedufn


src/lib.rs
```diff
@@ -370,6 +370,20 @@ impl CoreAnimationLayerRef {
     pub fn set_framebuffer_only(&self, framebuffer_only: BOOL) {
         unsafe { msg_send![self, setFramebufferOnly: framebuffer_only] }
     }
+
+    pub fn is_opaque(&self) -> bool {
```
chinedufn, line 374: Sweet!!

PR #170, opened 2020-08-11
Renamed CoreAnimationLayer to MetalLayer

adamnemecek:


examples/circle/main.rs
```diff
@@ -58,10 +58,10 @@ fn main() {
     // Set the command queue used to pass commands to the device.
     let command_queue = device.new_command_queue();
 
-    // Currently, CoreAnimationLayer is the only interface that provide
+    // Currently, MetalLayer is the only interface that provide
     // layers to carry drawable texture from GPU rendaring through metal
     // library to viewable windows.
-    let layer = CoreAnimationLayer::new();
+    let layer = MetalLayer::new();
```
chinedufn, line 64: Small bike shedding question -> `CAMetalLayer`?

Inspired by -> https://developer.apple.com/documentation/quartzcore/cametallayer

adamnemecek: You don't want to have the same name as the objc class. We don't call MTLRenderCommandEncoder that, we cut the MTL.

Issue #178, opened 2020-09-02
Synchronizing events


mxpv:
Hello! Currently `metal-rs` provides no APIs to synchronize GPU/CPU work (like [MTLEvent](https://developer.apple.com/documentation/metal/mtlevent) and [MTLSharedEvent](https://developer.apple.com/documentation/metal/mtlsharedevent), example: [Synchronizing Events Within a Single Device](https://developer.apple.com/documentation/metal/synchronization/synchronizing_events_within_a_single_device)).
Would it be ok to open a PR to support that?

kvark:
Yes, it would be great to have!

PR #179, opened 2020-09-03
Add sync primitives 

mxpv:
This PR adds Metal's sync primitives - Event, SharedEvent, Fence.
`SharedEvent.notify` stuff is a bit messy, but I couldn't find a better way.

ref: https://github.com/gfx-rs/metal-rs/issues/178

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,9 @@ name = "bind"
 [[example]]
 name = "events"
 
+[[example]]
+name = "fence"
+
 [workspace]
 members = [
   "examples/texture",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "metal"
-version = "0.20.0"
+version = "0.20.1"
 description = "Rust bindings for Metal"
 documentation = "https://docs.rs/crate/metal"
 homepage = "https://github.com/gfx-rs/metal-rs"
@@ -18,6 +18,7 @@ default-target = "x86_64-apple-darwin"
 [features]
 default = []
 private = []
+dispatch-queue = ["dispatch"]
 
 [dependencies]
 cocoa-foundation = "0.1"
@@ -25,6 +26,7 @@ bitflags = "1"
 log = "0.4"
 block = "0.1.5"
 foreign-types = "0.3"
+dispatch = { version = "0.2.0", optional = true }
 
 [dependencies.objc]
 version = "0.2.4"
@@ -76,6 +78,7 @@ name = "bind"
 
 [[example]]
 name = "events"
+required-features = ["dispatch-queue"]
 
 [[example]]
 name = "fence"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,9 @@ path = "examples/compute/compute-argument-buffer.rs"
 [[example]]
 name = "bind"
 
+[[example]]
+name = "events"
+
 [workspace]
 members = [
   "examples/texture",
--- a/src/device.rs
+++ b/src/device.rs
@@ -1372,12 +1372,7 @@ extern "C" {
         queue: dispatch_queue_t,
         destructor: dispatch_block_t,
     ) -> dispatch_data_t;
-    pub fn dispatch_release(object: dispatch_data_t); // actually dispatch_object_t
-
-    pub fn dispatch_queue_create(
-        label: *const std::os::raw::c_char,
-        attr: *const std::ffi::c_void,
-    ) -> dispatch_queue_t;
+    fn dispatch_release(object: dispatch_data_t); // actually dispatch_object_t
 }
 
 /*type MTLNewLibraryCompletionHandler = extern fn(library: id, error: id);
--- a/src/device.rs
+++ b/src/device.rs
@@ -1731,6 +1731,10 @@ impl DeviceRef {
         unsafe { msg_send![self, newSharedEvent] }
     }
 
+    pub fn new_fence(&self) -> Fence {
+        unsafe { msg_send![self, newFence] }
+    }
+
     pub fn heap_buffer_size_and_align(
         &self,
         length: NSUInteger,
--- a/src/device.rs
+++ b/src/device.rs
@@ -1350,7 +1350,7 @@ extern "C" {
 #[allow(non_camel_case_types)]
 type dispatch_data_t = id;
 #[allow(non_camel_case_types)]
-type dispatch_queue_t = id;
+pub type dispatch_queue_t = id;
 #[allow(non_camel_case_types)]
 type dispatch_block_t = *const Block<(), ()>;
 
@@ -1372,7 +1372,12 @@ extern "C" {
         queue: dispatch_queue_t,
         destructor: dispatch_block_t,
     ) -> dispatch_data_t;
-    fn dispatch_release(object: dispatch_data_t); // actually dispatch_object_t
+    pub fn dispatch_release(object: dispatch_data_t); // actually dispatch_object_t
+
+    pub fn dispatch_queue_create(
+        label: *const std::os::raw::c_char,
+        attr: *const std::ffi::c_void,
+    ) -> dispatch_queue_t;
 }
 
 /*type MTLNewLibraryCompletionHandler = extern fn(library: id, error: id);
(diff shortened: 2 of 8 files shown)


src/sync.rs
```diff
@@ -0,0 +1,133 @@
...
+            >(block);
+            (*block).flags |= BLOCK_HAS_SIGNATURE | BLOCK_HAS_COPY_DISPOSE;
+            (*block).extra = &BLOCK_EXTRA;
+            let () = msg_send![self, notifyListener:listener atValue:value block:block];
+            mem::forget(block);
+        }
+
+        extern "C" fn dtor(_: *mut BlockBase<(&SharedEventRef, u64), ()>) {}
+
+        static mut SIGNATURE: &[u8] = b"v16@?0Q8\0";
```
chinedufn, line 67: Why static instead of `const` on these?

chinedufn (approved):
Thanks!!!

mxpv:
@chinedufn updated

src/device.rs
```diff
@@ -1350,7 +1350,7 @@ extern "C" {
 #[allow(non_camel_case_types)]
 type dispatch_data_t = id;
 #[allow(non_camel_case_types)]
-type dispatch_queue_t = id;
+pub type dispatch_queue_t = id;
```
kvark, line 1353: I think it doesn't exactly fit the rest of the API that these types and C functions are exported. Can we instead do this properly with a new type responsible for the dispatch queue?

kvark (commented):
Great stuff in here, thank you!

src/device.rs
mxpv: Can you elaborate on this? There are also `dispatch_queue_create` and `dispatch_release` that has to be called in order to create a queue (and `dispatch_queue_create`  is also used for library stuff).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "metal"
-version = "0.20.0"
+version = "0.21.0"
```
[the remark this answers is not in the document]
mxpv: oh, sorry. I thought that `minor` is for API changes. will change to `0.20.1`

src/device.rs
kvark: The way dispatch queues are exposed with this PR is inconsistent with other entry points. For example, all the label API operates on `&str`, and the library is responsible for conversion. Also, it has reference counting built in. So with all that said, the dispatch queues should probably be exposed by some opaque struct type:
```rust
pub struct DispatchQueue {... }
impl DispatchQueue {
  pub fn new() -> Self;
  ...
}
```

src/sync.rs
```diff
@@ -43,6 +43,11 @@ impl SharedEventRef {
         unsafe { msg_send![self, setSignaledValue: new_value] }
     }
 
+    /// Schedules a notification handler to be called after the shareable event’s signal value
+    /// equals or exceeds a given value.
+    ///
+    /// Note: internally this function will append `BLOCK_HAS_SIGNATURE` and `BLOCK_HAS_COPY_DISPOSE`
```
chinedufn, line 49: Oh my apologies, I didn't realize that this was just an internal detail that the user didn't need to know about!

I misread it as it being possible for this function to segfault.

src/device.rs
mxpv: Yeah, I understand what you mean. My thinking was is that queue management is beyond the scope of this crate, and ideally we should not expose any related API. So I exposed as less as possible (and leave it up to user how to approach). 
In servo they use `dispatch` crate: https://github.com/servo/pathfinder/blob/master/metal/src/lib.rs#L175, which is good. So should we just pull in this dependency instead of writing own wrappers?
(Personally I'd prefer to leave it out and accept a ptr to `dispatch_queue_t` in `from_dispatch_queue(queue: dispatch_queue_t)`. Users can decide whether they want to use ffi calls and pull in dispatch crate.

kvark: Ok. Would it be reasonable to have `dispatch` as an optional dependency?
I don't have any other suggestions, and I'm fine with landing the code as is otherwise.

mxpv: Having optional dependency sounds reasonable to me. Updated, PTAL.

mxpv:
@chinedufn @kvark Addressed comments, PTAL. 
With optional `dispatch` dependency events example looks more sane :)

Cargo.toml
```diff
@@ -16,15 +16,17 @@ exclude = ["guide/**/*", "examples/texture/**/*", "tests/**/*", "Cargo.lock", "t
 default-target = "x86_64-apple-darwin"
 
 [features]
-default = []
+default = ["dispatch_queue"]
```
kvark, line 19: let's not have this as a default, please

src/device.rs
kvark, line 1353: does this still need to be exposed like this, given that we agreed on having the `dispatch` dependency?

Cargo.toml
```diff
@@ -16,15 +16,17 @@ exclude = ["guide/**/*", "examples/texture/**/*", "tests/**/*", "Cargo.lock", "t
...
 private = []
+dispatch_queue = ["dispatch"]
```
kvark, line 21: do we need this feature specifically called `dispatch_queue`? Why not just `dispatch`?

src/device.rs
mxpv: There will be no way to create a SharedEventListener when `dispatch` feature is off. I guess we should allow users to use their own implementations and pass `dispatch_queue_t` handle.

Cargo.toml
mxpv: `dispatch` is sort of vague word and can be used in many contexts (like dispatch compute).
But I can rename it if you want.

Cargo.toml
mxpv: This will make SharedEventListener (and hence SharedEvent) not to usable by default. Is there a reason not to have this in by default? (I understand that having extra dependency sucks, but sync primitives is quite useful part of metal).
Are you sure?

src/device.rs
mxpv: It wasn't needed because it wasn't used in public API at all.
The type itself is not important, it's more like whether we allow to create `SharedEventListener` from raw handle (see [here](https://github.com/gfx-rs/metal-rs/pull/179/files#diff-2c2b428f563998bbe0d60c83167ea631R93)) or the only way would be via turning on the `dispatch_queue` switch and pulling in extra dependency.

kvark (approved):

kvark:
I think the last thing missing is fixing the default features.

mxpv:
@kvark Done!

mxpv:
CI tests need `--all-features` as well. Added in https://github.com/gfx-rs/metal-rs/pull/179/commits/238e79244bca0dd660368275ccdf9f319c4e7eba

kvark:
`cargo test --all` should also pass without all features,  or in any combination of features, in fact.
The last commit doesn't address the actual problem - the code which was using `dispatch` dependency wasn't properly feature-gated. Could you revert the last commit and instead attempt a proper fix? We can discuss the CI change separately in another PR.

Also, now it's not clear which commits are even buildable and testable on their own, so we'll have to squash it. Are you ok with this?

mxpv:
> the code which was using dispatch dependency wasn't properly feature-gated.

Not exactly. CI failed with:
```
error[E0432]: unresolved import `dispatch`
##[error] --> examples/events/main.rs:8:5
```
since we disabled `dispatch` feature by default. And [currently there is no way](https://github.com/rust-lang/cargo/issues/4663) to enable a feature per example.

kvark:
Why isn't this a required feature for the `events` example?

mxpv:
> Why isn't this a required feature for the events example?

`events` example does:
```rust
    // Shareable event listener
    let my_queue = Queue::create(
        "com.example.apple-samplecode.MyQueue",
        QueueAttribute::Serial,
    );

    let shared_event_listener = SharedEventListener::from_queue(&my_queue);
```

which requires `dispatch_queue`  to be enabled in order to call `from_queue` (and which is not enabled by default).
So when we run `cargo test --all`, it fails.

kvark:
We need `required-features = ["dispatch_queue"]` to be enabled for this example. This way, it will not be built by `cargo test` if the feature is not enabled.
Speaking of which, we should probably rename it to `dispatch-queue` (notice the dash)

mxpv:
crap, I misunderstood the doc. Yep, `required-features` did the trick. Should be good now.
Squashed the commits.

kvark (approved):
bors r+

PR #184, opened 2020-10-01
Remove NSRange from public api

Gordon-F:


chinedufn:
Since in Rust a `Range` is typically a `start..end`, replacing `NSRange` (documented as start, length) with a tuple might lead to confusion.

kvark:
A few questions:
  1. what's the motivation for removing it?
  2. what's the reason `ops::Range` wasn't used instead?

Gordon-F:
@chinedufn Thanks! Agree with you.

@kvark I think it would be nice to remove external dependency from public api. How about to inline cocoa-foundation types? Like we did with core-graphics (#157).

Also this changes can improve compile time a little bit.

kvark:
`cocoa-foundation` is a recent crate created specifically as a lightweight type-only dependency. So it shouldn't be a problem to use, and it will unlikely contribute negatively to compile times (unless you have some numbers showing the contrary?).

I guess the question is - would hiding `NSRange` let us exclude it from the public dependencies?

Gordon-F:
> unless you have some numbers showing the contrary

Some experiments:
```
rm -rf ./target && RUSTC_WRAPPER="" cargo +nightly build -j 4 -Z timings
```

```master (0c74e22):``` 9.2s
```without cocoa-foundation:``` 8.4s

kvark:
Thanks! And, to clarify, you expect the users of `metal-rs` to not depend on `cocoa-foundation` either? For example, `cocoa` depends on it.

Gordon-F:
> you expect the users of `metal-rs` to not depend on `cocoa-foundation` either

Yes.

> For example, `cocoa` depends on it.

But currently `metal-rs` use `cocoa` as dev-dependency. And i expect that users of `metal-rs` can add only `metal-rs` into dependencies and use it.

What we can do:
1. Nothing. `metal-rs` users add same version of `cocoa-foundation` dependency (we already have one in dependency tree)
2. Remove `cocoa-foundation` types from public API
3. Inline `cocoa-foundation` types and export it (also we can delete `cocoa-foundation` dependency in `gfx`).

kvark:
@Gordon-F would you have an example somewhere that wouldn't depend on `cocoa`? I.e. how do you create a window and stuff?

Gordon-F:
@kvark You mean `cocoa-rs` ? Any C/C++ based libs. Like `rust-sdl2` (`SDL`) or `miniquad` (`sokol-app`).

kvark:
I agree with @chinedufn that tuples look confusing from the API point of view. Rust ranges would be clear, but they require translation to/from NSRange :/ Maybe it's ok? Just one addition/subtraction.

chinedufn:
I'm thinking it's best to use `start, length` in order to avoid adding API use friction (mentally converting between NSRange and Range when you're figuring out what you need to pass to Metal).

So .. I think we can just do the same thing that we did in #157 

In #157 we replaced `core_graphics::geometry::CGSize` with our own `CGSize` struct.

Here we can just make our own `NSRange` struct.

If in the future anyone needs compatibility with `core_foundation` for whatever reason .. we can consider adding a feature flag.

But right now we have no reason to believe that maintaining compatibility is useful (as far as I know) - so *let's replace `NSRange` with our own `pub struct NSRange`*.

Gordon-F:
@chinedufn Done

kvark (approved):

PR #185, opened 2020-10-03
added CAMetalLayer methods

adamnemecek:


src/lib.rs
```diff
@@ -451,13 +469,13 @@ pub use {
 pub use mps::*;
 
 #[inline]
-unsafe fn obj_drop<T>(p: *mut T) {
```
kvark, line 454: these should probably be left unsafe

src/library.rs
```diff
@@ -121,13 +121,13 @@ impl FunctionConstantValues {
 }
 
 impl FunctionConstantValuesRef {
-    pub unsafe fn set_constant_value_at_index(
+    pub fn set_constant_value_at_index(
```
kvark, line 124: this can't be safe, since we are passing a raw pointer in

src/pipeline/render.rs
```diff
@@ -163,26 +163,28 @@ foreign_obj_type! {
 
 impl RenderPipelineReflection {
     #[cfg(feature = "private")]
-    pub unsafe fn new(
+    pub fn new(
```
kvark, line 166: similarly, can't be safe

kvark (changes requested):
We need to go through the methods and come up with a sound explanation as to why some of them are unsafe but others are not.

PR #189, opened 2020-12-03
Various missing bits and bobs

bsekura:
Sorry for dubious description. 
This MR is result of me working with metal-rs on some rendering code and adding missing API elements as I needed them.
I will continue this as I find the time, and in the meantime would appreciate considering merging what I got so far.

Feel free to ask me to cleanup some things (I ran into merge conflicts, as some of the functionality was being added by your team at the same time). I noticed some inconsistencies with enum naming and was not sure whether to follow Apple's or yours.

Thanks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,8 @@ mps = []
 cocoa-foundation = "0.1"
 bitflags = "1"
 log = "0.4"
-block = "0.1.5"
-foreign-types = "0.3"
+block = "0.1.6"
+foreign-types = "0.3.2"
 dispatch = { version = "0.2.0", optional = true }
 
 [dependencies.objc]
@@ -34,7 +34,7 @@ version = "0.2.4"
 features = ["objc_exception"]
 
 [dev-dependencies]
-cocoa = "0.23"
+cocoa = "0.24.0"
 cty = "0.2.1"
 winit = "0.22"
 sema = "0.1.4"
--- a/src/device.rs
+++ b/src/device.rs
@@ -54,13 +54,31 @@ pub enum MTLFeatureSet {
     macOS_GPUFamily2_v1 = 10005,
 }
 
+#[repr(i64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLGPUFamily {
+    Common1 = 3001,
+    Common2 = 3002,
+    Common3 = 3003,
+    Apple1 = 1001,
+    Apple2 = 1002,
+    Apple3 = 1003,
+    Apple4 = 1004,
+    Apple5 = 1005,
+    Apple6 = 1006,
+    Mac1 = 2001,
+    Mac2 = 2002,
+    MacCatalyst1 = 4001,
+    MacCatalyst2 = 4002,
+}
+
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLDeviceLocation {
-    MTLDeviceLocationBuiltIn = 0,
-    MTLDeviceLocationSlot = 1,
-    MTLDeviceLocationExternal = 2,
-    MTLDeviceLocationUnspecified = u64::MAX,
+    BuiltIn = 0,
+    Slot = 1,
+    External = 2,
+    Unspecified = u64::MAX,
 }
 
 bitflags! {
@@ -1334,12 +1352,29 @@ impl MTLFeatureSet {
     }
 }
 
-#[allow(non_camel_case_types)]
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLArgumentBuffersTier {
-    tier1 = 0,
-    tier2 = 1,
+    Tier1 = 0,
+    Tier2 = 1,
+}
+
+#[repr(u64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLReadWriteTextureTier {
+    TierNone = 0,
+    Tier1 = 1,
+    Tier2 = 2,
+}
+
+#[repr(u64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLCounterSamplingPoint {
+    AtStageBoundary = 0,
+    AtDrawBoundary = 1,
+    AtDispatchBoundary = 2,
+    AtTileDispatchBoundary = 3,
+    AtBlitBoundary = 4,
 }
 
 bitflags! {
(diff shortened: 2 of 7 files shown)


src/device.rs
```diff
@@ -1334,12 +1353,29 @@ impl MTLFeatureSet {
     }
 }
 
-#[allow(non_camel_case_types)]
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLArgumentBuffersTier {
-    tier1 = 0,
```
chinedufn, line 1341: What's your thought process here for making capital?

src/device.rs
```diff
@@ -54,6 +54,25 @@ pub enum MTLFeatureSet {
     macOS_GPUFamily2_v1 = 10005,
 }
 
+#[allow(non_camel_case_types)]
+#[repr(i64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLGPUFamily {
+    common1 = 3001,
```
chinedufn, line 61: What's your thought process here for making lowercase?

chinedufn (commented):
Awesome.

Left a small question. 

Looks like we'd all benefit from creating docs for naming convention guidelines at some point.

src/device.rs
```diff
@@ -1732,16 +1841,28 @@ impl DeviceRef {
...
             } else {
                 None
             }
         }
     }
 
+    pub fn read_write_texture_support(&self) -> Option<MTLReadWriteTextureTier> {
+        unsafe {
+            let has_selector: BOOL =
+                msg_send![self, respondsToSelector: sel!(readWriteTextureSupport)];
```
kvark, line 1857: Let's have a discussion about these checks. I think we need to be consistent there:
  1. either we check for all the selectors that are potentially unsupported
  2. or we don't check at all, but document that a function requires particular GPU family supported

Metal is an inherently unsafe API. Just because you can use a selector doesn't mean that it's valid at some point to use it. So these checks aren't going to make the functions safe, anyway. Users still need to ensure that they have the proper context, and they are calling it right. With this in mind, I think we should go with (2) and strip away the `respondsToSelector` checks entirely. Instead, we should try putting proper doc comments about when a function is supported.

kvark (commented):
This looks wonderful, thank you!
In addition to @chinedufn notes I have a concern about `respondsToSelector` that I'd like to resolve before merging.

src/device.rs
bsekura: Just following existing enums naming convention.

Let me take this opportunity to ask about enum naming convention. I have noticed some small inconsistencies, e.g. `MTLDeviceLocation`, that repeats the type name inside variant name. Perhaps it was a hasty copy pasta from objc headers.

Looking at original Apple stuff, we have objc:

```objc
typedef enum MTLDeviceLocation : NSUInteger {
    MTLDeviceLocationBuiltIn,
    MTLDeviceLocationSlot
} MTLDeviceLocation;
```

And Swift:

```swift
enum MTLDeviceLocation : UInt {
    case builtIn
    case slot
}
```

For Rust, I suggest to simply take Swift declaration and simply capitalize the first word.
I think this what you guys have intended, so I'm just confirming.

I can fix `MTLDeviceLocation` as part of this MR if you wish.

src/device.rs
bsekura: That's a mistake, I confused myself with `MTLFeatureSet` starting with lowercase.
I will fix it.

Related to the comment above about enum conventions, `MTLFeatureSet` is a special case because the enums are product names that Apple decided to start with lowercase. I think that special case is fine.

src/device.rs
bsekura: I agree.

On the one hand, having selector checks and nicely wrapping the result in `Option` seems appealing, but then it's a pain in the ass as `metal-rs` developers have to keep track of which of these selectors depend on API version and change the signature. It's better to remove checks and let the user take responsibility in checking the documentation on their availability (query the OS version, etc.).

There are a few existing methods that were doing that check before my MR, e.g. `argument_buffers_support`.
Would you like me to remove them all as part of this MR?

src/device.rs
chinedufn: I think the inconsistencies come from different folks adding in different types using different styles.

> I can fix MTLDeviceLocation as part of this MR if you wish.

Let's hold off for now and get what you have landed. But, if you just so happen to want to open a separate issue to nail down naming conventions that we can begin to enforce going forwards - that would be fantastic.

src/device.rs
kvark: If there is a few outliers, then yes, let's just get them cleaned up. Otherwise, we can follow-up.

kvark (approved):

Cargo.toml
```diff
@@ -25,16 +25,16 @@ mps = []
 cocoa-foundation = "0.1"
 bitflags = "1"
 log = "0.4"
-block = "0.1.5"
-foreign-types = "0.3"
+block = "0.1.6"
```
kvark, line 28: is there a particular reason for these deps to be bumped? i.e. does any of the new code require this?

src/device.rs
```diff
@@ -1484,12 +1531,19 @@ impl DeviceRef {
...
 
-    pub fn location(&self) -> MTLDeviceLocation {
-        unsafe { msg_send![self, location] }
-    }
-
-    pub fn location_number(&self) -> NSUInteger {
-        unsafe { msg_send![self, locationNumber] }
+    pub fn supports_raytracing(&self) -> bool {
+        unsafe {
+            let has_sel: BOOL = msg_send![self, respondsToSelector: sel!(supportsRaytracing)];
```
kvark, line 1536: let's remove this check?

kvark:
Please squash the commits before landing :)

Cargo.toml
bsekura: Figured I update point release for potential bug fixes.

kvark: The libraries need to point to dependency versions that they require. It's then up to the final applications (or dynamic libraries, basically - all binary targets) to carry `Cargo.lock` and update all these revisions as they please.

src/device.rs
```diff
@@ -1730,16 +1833,12 @@ impl DeviceRef {
         unsafe { msg_send![self, newDepthStencilStateWithDescriptor: descriptor] }
     }
 
-    pub fn argument_buffers_support(&self) -> Option<MTLArgumentBuffersTier> {
-        unsafe {
-            let has_arg_buffers: BOOL =
```
kvark, line 1735: `BOOL` is now unused. Please check the warnings!

kvark (commented):
Looks amazing, thank you!
One last tiny bit of a thing :)

Cargo.toml
bsekura: Looks like I don't quite understand how dependency versions work across libraries and apps that use them ;)
Would you like me to revert it?

kvark: Let me try to explain some more. We don't want to constantly try to keep every dependency to the latest version. We want to have the latest *breaking* version only, i.e. "0.1" for this `block` would be fine to keep. The reason it was "0.1.5" instead is likely because a particular change in metal-rs required something that was added in `block-0.1.5` but was not in 0.1.4. So, if somebody tries to build the library with *minimal* dependency versions, it would still work.
This way, we can check the `git blame` for this dependency line and see who and why bumped it.

If we just blindly bump it every time we touch the code, then:
  - we are going to have more churn in the code
  - all the library clients would have to build the updated `block` dependency once they update `metal-rs`. This may cause problem in case, of a higher block version having a regression in performance, for example.
  - we no longer know what's the actual *minimum* version of that dependency required

TL;DR: only bump the non-breaking version if it's required, never without reason

bsekura: Thank you for explanation!

bsekura:
This is probably a wrong place to ask, but I will be adding the new raytracing API and perhaps a small example. Just wanted to  confirm if you guys also plan working on this, to avoid duplicating effort. If not, I'll proceed and prepare PR in the future.

kvark:
Yeah, that sounds awesome!

PR #195, opened 2021-01-12
window example: Fix rendering on hdpi screens

xixixao:
Without this PR:

Screen Shot 2021-01-11 at 23 47 10

After this PR:

Screen Shot 2021-01-11 at 23 47 48

I haven't tested the change of scale factor, but it looks reasonable :)


kvark (approved):
thank you!

PR #196, opened 2021-01-15
Prevent SIGSEGV when returning errors from new_library_with_source

xixixao:
This call SIGSEGVs on my system (10.15.7). Why? And will this leak memory? I have no idea.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1688,7 +1688,6 @@ impl DeviceRef {
                 let compile_error: *const std::os::raw::c_char = msg_send![desc, UTF8String];
                 let message = CStr::from_ptr(compile_error).to_string_lossy().into_owned();
                 if library.is_null() {
-                    let () = msg_send![err, release];
                     return Err(message);
                 } else {
                     warn!("Shader warnings: {}", message);


chinedufn:
I searched the files commit history and found -> https://github.com/gfx-rs/metal-rs/commit/d8977de5873be7b7adce13a8d93433471a983b80#

kvark:
In addition, could you post the full stack for the SIGSEGV? I'm interested in knowing if it happens on `release` or some time later.

xixixao:
Ok, this might be helpful. The sigsegv happens when I do this inside `autorelasepool`. You can run this example to repro:

```rust
use metal::*;
use objc::rc::autoreleasepool;

fn main() {
  let device = Device::system_default().expect("no device found");
  let shader_source = "function (";
  autoreleasepool(|| {
    let library = device.new_library_with_source(shader_source, &CompileOptions::new());
    eprintln!("{}", library.is_ok());
  });
}
```

I pushed it to this branch:
https://github.com/xixixao/metal-rs/tree/show-new-library-sigsegv

Am I just doing things wrong? If yes, how do I know what I can and cannot call inside `autoreleasepool` (which is used inside the `window` example)?

xixixao:
I'm lost. It crashes at the end of the autoreleasepool. I'm not sure why it only crashes on this particular object. Maybe something with the way the NSError is created inside `newLibraryWithSource` (but I guess Metal is not open source so I can't find its source code anywhere).

xixixao:
I think this is the right approach: https://github.com/gfx-rs/metal-rs/pull/198

chinedufn:
@xixixao thank you for investigating this!

kvark (approved):

Issue #40, opened 2018-05-03
RFC: Send+Sync on metal objects


kvark:
Given that all the interaction is done via message sending, I don't see why Send+Sync wouldn't be the case?

kvark:
### Proposal
  1. Have `foreign_obj_type` forcefully implement `Send` + `Sync`
  2. Mark all the methods (non-static, corresponding to ObjC messages) with `&mut self`
  3. Mark atomic property get/set accessors with `&self`
  4. Mark non-atomic properties with `&mut self`.

### Clarification

Atomicity of properties in ObjC guarantees no visible partial writes. While Apple docs say this is not thread safe, they mean data-races in fact, and it's perfectly thread-safe in Rust terms (of memory safety).

kvark:
@gfx-rs/metallists 
Comments? Questions? Objections?
This is an important thing for us to decide.

kvark:
Sorry, I realized why this isn't going to work :disappointed: . `&mut self` means nothing if we are able to have multiple independent rust objects corresponding to the same ObjC object. There has to be something our Rust objects share between each other, e.g. all being `Arc<RwLock<ObjcClassRef>>` but with `Arc` replaced by something that uses native `ObjC` runtime counters.

kvark:
## Current approach

A bit of context here, that's what we currently have for ObjC bindings of Metal:
  - a `SomeClassRef` type with actual methods (immutable!) for properties and ObjC messages. It's a type-system thing in Rust that doesn't affect any ObjC reference counts. Type system guarantees that the object is alive for the lifetime of our XXXRef struct.
  - a `SomeClass` type that represent a strong reference to ObjC object. Cloning it is akin cloning an `Arc`. One can borrow it and get `SomeClassRef` temporarily.

## Analysis

The main problem for getting any sort of type assistance here is that `SomeClass` objects don't share anything with each other, so we don't know (even at run-time) if those point to the same underlying ObjC object...

So here comes the main idea and restriction: let's enforce a guarantee that there is a single Rust object backing a single ObjC object. On Rust side, there is no point to mess with ObjC reference counters. Instead, we can use the regular `Arc` and locking primitives. As long as ObjC runtime is concerned, we only have one object.

## Unresolved Problems

Any methods that return new objects to use (`newXXX`) would just return `SomeClass` by value, and the client code would then be responsible for wrapping it into sharing structures, if needed. The destructor of `SomeClass` would then release the only ObjC count.

[...]

PR #197, opened 2021-01-16
Various new features of macOS 11.0

bsekura:
This is another pass of adding new macOS 11.0 features to metal-rs. It lays some groundwork to add new ray tracing API, as well as adding things like dynamic libraries, function pointers, etc. Some naming changes were also made.

This is still work in progress, but I wanted to merge as I go to avoid drifting too much. There are some comments left for things yet to be done.

I also added a simple example of using shader dynamic libraries.

I expect some discussion and cleanup, so I will squash commits once we're all happy with this PR.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -9,7 +9,7 @@ use block::{Block, ConcreteBlock};
 use cocoa_foundation::base::id;
 use cocoa_foundation::foundation::NSUInteger;
 use foreign_types::ForeignType;
-use objc::runtime::{Object, BOOL, NO, YES};
+use objc::runtime::{Object, NO, YES};
 
 use super::*;
 
@@ -54,13 +54,31 @@ pub enum MTLFeatureSet {
     macOS_GPUFamily2_v1 = 10005,
 }
 
+#[repr(i64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLGPUFamily {
+    Common1 = 3001,
+    Common2 = 3002,
+    Common3 = 3003,
+    Apple1 = 1001,
+    Apple2 = 1002,
+    Apple3 = 1003,
+    Apple4 = 1004,
+    Apple5 = 1005,
+    Apple6 = 1006,
+    Mac1 = 2001,
+    Mac2 = 2002,
+    MacCatalyst1 = 4001,
+    MacCatalyst2 = 4002,
+}
+
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLDeviceLocation {
-    MTLDeviceLocationBuiltIn = 0,
-    MTLDeviceLocationSlot = 1,
-    MTLDeviceLocationExternal = 2,
-    MTLDeviceLocationUnspecified = u64::MAX,
+    BuiltIn = 0,
+    Slot = 1,
+    External = 2,
+    Unspecified = u64::MAX,
 }
 
 bitflags! {
@@ -1334,19 +1352,56 @@ impl MTLFeatureSet {
     }
 }
 
-#[allow(non_camel_case_types)]
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLArgumentBuffersTier {
-    tier1 = 0,
-    tier2 = 1,
+    Tier1 = 0,
+    Tier2 = 1,
+}
+
+#[repr(u64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLReadWriteTextureTier {
+    TierNone = 0,
+    Tier1 = 1,
+    Tier2 = 2,
+}
+
+/// Only available on (macos(11.0), ios(14.0))
+#[repr(u64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLCounterSamplingPoint {
+    AtStageBoundary = 0,
+    AtDrawBoundary = 1,
+    AtDispatchBoundary = 2,
+    AtTileDispatchBoundary = 3,
+    AtBlitBoundary = 4,
+}
+
+/// Only available on (macos(11.0), macCatalyst(14.0), ios(13.0))
+#[repr(u64)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum MTLSparseTextureRegionAlignmentMode {
+    Outward = 0,
+    Inward = 1,
 }
 
 bitflags! {
     struct MTLPipelineOption: NSUInteger {
-        const ArgumentInfo   = 1 << 0;
-        const BufferTypeInfo = 1 << 1;
+        const None                      = 0;
+        const ArgumentInfo              = 1 << 0;
+        const BufferTypeInfo            = 1 << 1;
+        /// Only available on (macos(11.0), ios(14.0))
+        const FailOnBinaryArchiveMiss   = 1 << 2;
     }
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
+#[repr(C)]
+pub struct MTLAccelerationStructureSizes {
+    pub acceleration_structure_size: NSUInteger,
+    pub build_scratch_buffer_size: NSUInteger,
+    pub refit_scratch_buffer_size: NSUInteger,
 }
 
 #[link(name = "Metal", kind = "framework")]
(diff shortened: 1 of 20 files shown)


src/device.rs
```diff
@@ -1377,13 +1378,32 @@ pub enum MTLCounterSamplingPoint {
...
+        const BufferTypeInfo            = 1 << 1;
+        /// Only available on (macos(11.0), ios(14.0))
+        const FailOnBinaryArchiveMiss   = 1 << 2;
     }
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
+#[repr(C)]
+pub struct MTLAccelerationStructureSizes {
+    acceleration_structure_size: NSUInteger,
```
kvark, line 1402: all these fields need to be public

src/heap.rs
```diff
@@ -107,4 +191,26 @@ impl HeapDescriptorRef {
     pub fn set_size(&self, size: NSUInteger) {
         unsafe { msg_send![self, setSize: size] }
     }
+
+    /// Only available on macos(10.15), ios(13.0)
+    pub fn hazard_tracking_mode(&self) -> MTLHazardTrackingMode {
+        unsafe { msg_send![self, hazardTrackingMode] }
+    }
+    /// Only available on macos(10.15), ios(13.0)
+    pub fn set_hazard_tracking_mode(&self, mode: MTLHazardTrackingMode) {
```
kvark, line 200: looks like the property is read-only, we can't have the setter

src/heap.rs
```diff
@@ -107,4 +191,26 @@ impl HeapDescriptorRef {
...
+        unsafe { msg_send![self, setHazardTrackingMode: mode] }
+    }
+
+    /// Only available on macos(10.15), ios(13.0)
+    pub fn resource_options(&self) -> MTLResourceOptions {
+        unsafe { msg_send![self, resourceOptions] }
+    }
+    /// Only available on macos(10.15), ios(13.0)
+    pub fn set_resource_options(&self, options: MTLResourceOptions) {
```
kvark, line 209: same here, we can't have the setter

src/library.rs
```diff
@@ -45,6 +56,86 @@ impl VertexAttributeRef {
             }
         }
     }
+
+    /// Only available on (macos(10.12), ios(10.0)
```
kvark, line 60: These comments are wonderful!

src/library.rs
```diff
@@ -64,6 +310,26 @@ foreign_obj_type! {
 }
 
 impl FunctionRef {
```
kvark, line 312: perhaps, the function and argument stuff could be moved out into something like `function.rs`?

src/library.rs
```diff
@@ -122,16 +409,45 @@ impl FunctionConstantValues {
 }
 
 impl FunctionConstantValuesRef {
-    pub unsafe fn set_constant_value_at_index(
+    pub fn set_constant_value_at_index(
         &self,
+        value: *const std::os::raw::c_void,
```
kvark, line 414: let's have the `c_void` included at the top of the module, just to be consistent

src/library.rs
```diff
@@ -172,23 +488,110 @@ impl CompileOptionsRef {
...
     }
+
+    /// Only available on (macos(11.0), ios(14.0))
+    pub fn library_type(&self) -> MTLLibraryType {
+        unsafe { msg_send![self, libraryType] }
+    }
+
+    /// Only available on (macos(11.0), ios(14.0))
+    pub fn set_library_type(&self, lib_type: MTLLibraryType) {
+        unsafe { msg_send![self, setLibraryType: lib_type] }
```
kvark, line 508: Note that there is a data race hazard here: the property is non-atomic. We discussed it in #40 but haven't come up with a solid plan. Marking the whole function as `&mut self` is overprotective (doesn't let you use other properties), and not effective enough at the same time (you can obtain multiple references to the same object). Sot it's just something to keep in mind if you are looking for a way to make this API better in Rust.

src/library.rs
```diff
@@ -172,23 +488,110 @@ impl CompileOptionsRef {
...
+        unsafe {
+            let libraries: *mut Object = msg_send![self, libraries];
+            let count: NSUInteger = msg_send![libraries, count];
+            let ret = (0..count)
+                .map(|i| {
+                    let lib = msg_send![libraries, objectAtIndex: i];
+                    DynamicLibrary::from_ptr(lib)
+                })
+                .collect();
+            let () = msg_send![libraries, release];
```
kvark, line 540: why does it need to be released? My understanding is that it's not a retained object here, it's borrowed from the start.

src/library.rs
```diff
@@ -257,4 +663,334 @@ impl LibraryRef {
...
+    /// Only available on (macos(11.0), ios(14.0))
+    pub fn library_type(&self) -> MTLLibraryType {
+        unsafe { msg_send![self, type] }
+    }
+
+    /// Only available on (macos(11.0), ios(14.0))
+    pub fn install_name(&self) -> Option<&str> {
+        unsafe {
+            let maybe_name: *mut Object = msg_send![self, installName];
+            if !maybe_name.is_null() {
```
kvark, line 676: nit: the `as_ref()` method works nice for cases like this, i.e. this would be `maybe_name.as_ref().map(crate::nsstring_as_str)`

src/library.rs
```diff
@@ -257,4 +663,334 @@ impl LibraryRef {
...
+        descriptor: &RenderPipelineDescriptorRef,
+    ) -> Result<bool, String> {
+        unsafe {
+            let mut err: *mut Object = ptr::null_mut();
+            let result: BOOL = msg_send![self, addRenderPipelineFunctionsWithDescriptor:descriptor
+                                                                        error:&mut err];
+            if !err.is_null() {
+                // FIXME: copy pasta
+                let desc: *mut Object = msg_send![err, localizedDescription];
+                let c_msg: *const std::os::raw::c_char = msg_send![desc, UTF8String];
```
kvark, line 887: we should probably include `c_char`, or at least the whole `std::os::raw`

src/pipeline/compute.rs
```diff
@@ -178,7 +290,14 @@ impl ComputePipelineStateRef {
         }
     }
 
-    pub fn max_total_threads_per_group(&self) -> NSUInteger {
+    pub fn set_label(&self, label: &str) {
```
kvark, line 293: we can't have that, the property is [read-only](https://developer.apple.com/documentation/metal/mtlcomputepipelinestate/2880323-label?language=objc)

kvark (commented):
Amazing work here, happy to see this growing!

src/library.rs
bsekura: OK, I'm gonna need some help understanding Objective C rules here.
I was following what `function_names` method is doing, and its objc signature is:

```objc
@property(readonly) NSArray<NSString *> *functionNames;
```

This guy's signature is:

```obj
@property(readwrite, nullable, copy, nonatomic) NSArray<id<MTLDynamicLibrary>> *libraries;
```

In both cases we marshal values to Rust types. Any advice?

src/library.rs
bsekura: Very nice, I will fix it.

src/library.rs
kvark: Perhaps, `function_names` is just wrong as well?
I'm not an Apple API expect. My mental model is simple. Objc documentation says (somewhere) that any calls to "newXxx" or "createXxx" return retained references, i.e. ones we "own". This one just returns a pointer to `NSArray`, it's an accessor, it doesn't retain anything. So the array is still owned by the same thing, we just get a reference to it. Therefore, we shouldn't be releasing it.
@scoopr could you check if we are missing anything here?

scoopr: Yes, the `release` looks wrong to me in both cases, just accessing a property shouldn't involve refcounting

bsekura: OK, cool, that makes sense. Now there may be other cases like this across the code base that needs fixing then. 
Let me stick to the scope of this PR and I will go ahead and make a pass to clean up these.

src/library.rs
bsekura: Thanks for the context. At this time I cannot offer a solution, but will keep this in mind in the future.

kvark:
Please use "Resolve Conversation" on items that are addressed, so that we know when it's time to look at the code again.

kvark:
@bsekura what's the status? Is it waiting for re-review of some points, or all ready?

src/library.rs
bsekura: Yeah, I was thinking about the same thing. However, there is still quite a few interfaces missing and I was planning on splitting files once I have everything so I can get my head around it and figure out the best grouping. Mind if we leave it for another PR?

kvark: don't mind at all!

src/library.rs
bsekura: In general, I try to follow the original apple header comments. There are some inconsistencies across the codebase and maybe it's good to create an issue to clean them all up at some point to improve the docs.

bsekura:
That should do it for now. Please check and merge if there are no issues. Sorry, I was unavailable for a bit and it took longer than expected.

kvark (commented):
One last thing!

bsekura:
I'm on it.
Squash via rebase is giving me a headache because of merge with upstream.
Are you able to merge and squash on your end using github's features?

kvark (approved):
Great, thank you!

PR #198, opened 2021-01-17
Prevent SIGSEGV when returning errors from new_library_with_source 2

xixixao:
Ok this is not cleaned up, but directionally I think the way to go. I think the NSError that gets created inside `newLibraryWithSource:source` is autoreleased. So to prevent other wrapping autoreleasepool code from seg faulting we always use autoreleasepool.

While I'm here, I have 1 more problem with this API:

- Warning is just printed - what if I want to use it? It would be better to return the warning along with the library.

### DIFF
--- a/examples/error-sigsegv/main.rs
+++ b/examples/error-sigsegv/main.rs
@@ -8,4 +8,5 @@ fn main() {
     let library = device.new_library_with_source(shader_source, &CompileOptions::new());
     eprintln!("{}", library.is_ok());
   });
+  eprintln!("the autorelease pool does not crash");
 }
--- a/examples/error-sigsegv/main.rs
+++ b/examples/error-sigsegv/main.rs
@@ -1,0 +1,11 @@
+use metal::*;
+use objc::rc::autoreleasepool;
+
+fn main() {
+  let device = Device::system_default().expect("no device found");
+  let shader_source = "function (";
+  autoreleasepool(|| {
+    let library = device.new_library_with_source(shader_source, &CompileOptions::new());
+    eprintln!("{}", library.is_ok());
+  });
+}
--- a/src/device.rs
+++ b/src/device.rs
@@ -9,6 +9,7 @@ use block::{Block, ConcreteBlock};
 use cocoa_foundation::base::id;
 use cocoa_foundation::foundation::NSUInteger;
 use foreign_types::ForeignType;
+use objc::rc::autoreleasepool;
 use objc::runtime::{Object, BOOL, NO, YES};
 
 use super::*;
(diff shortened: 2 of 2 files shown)


examples/error-sigsegv/main.rs
```diff
@@ -0,0 +1,12 @@
+use metal::*;
```
chinedufn, line 1: Maybe something like

```
#[cfg(tests)]
mod tests {
  // Verify that erroneous library source does not crash.
  #[test]
  fn autorelease_new_library_with_source_error() {
  }
}
```

At the bottom of device.rs?

src/device.rs
```diff
@@ -1676,23 +1677,28 @@ impl DeviceRef {
...
-            if !err.is_null() {
-                let desc: *mut Object = msg_send![err, localizedDescription];
-                let compile_error: *const std::os::raw::c_char = msg_send![desc, UTF8String];
-                let message = CStr::from_ptr(compile_error).to_string_lossy().into_owned();
-                if library.is_null() {
-                    let () = msg_send![err, release];
-                    return Err(message);
-                } else {
-                    warn!("Shader warnings: {}", message);
+            autoreleasepool(|| {
```
chinedufn, line 1683: I'll default to kvark on this since I don't yet know much about autoreleasepool.

kvark: We don't want to use `autoreleasepool` *inside* the method implementations here. It's a low level wrapper of Metal APIs, so we defer the management of autorelease pools to the user.

kvark (changes requested):
I'm confused about these changes. You first tried to remove the `release` on the error. That seems like the right fix. What was the problem with it?

xixixao:
If you're happy to remove the release from the method, that works, but it likely will leak memory. I haven't used the right memory tool, but you would likely find it does leak.

Why don't we want to use autoreleasepool inside the methods? I'm not 100% sure, but based on my investigation it seems that the `newLibraryWithSource` method expects to be called inside `autoreleasepool`. And afaik there is no harm in having nested `autoreleasepool`s.

kvark:
> If you're happy to remove the release from the method, that works, but it likely will leak memory

Let's confirm if it's the case then, before fixing!

> And afaik there is no harm in having nested autoreleasepools.

I think we should either use ARP everywhere, or nowhere. The latter seems to be better suited for a low-level API wrapper like metal-rs. I could imagine a case where he user wants to minimize the amount of work at the time `newLibraryWithSource` is called (to reduce the frame hitches, for example)? They'd have an autorelease pool freed some time after the frame is done, instead of in the middle of it.

xixixao:
> Let's confirm if it's the case then, before fixing!

I don't know how to do this, do you? (but I also don't see how it would not leak memory - how would the retainCount get to 0?)

> I think we should either use ARP everywhere, or nowhere.

My understanding is we don't have a choice. If the Metal framework calls `[err autorelease]` inside `newLibraryWithSource`, then if we don't include `autoreleasepool` we will leak memory any time the caller of the rust method doesn't use `autoreleasepool` themselves. Now it is bizarre way of structure the Apple API, and I tried to find a way to detect whether there's an active `autoreleasepool` but I didn't find one.

Unless you're suggesting that all examples should include `autoreleasepool` wrapping the whole `main` method?

xixixao:
Also

>  I could imagine a case where he user wants to minimize the amount of work at the time

The `autoreleasepool` only causes the 2 releases that were already called explicitly in the existing code (the string and err). So it seems that if we were ok with the original code we'd be ok with this one.

grovesNL:
> Unless you're suggesting that all examples should include `autoreleasepool` wrapping the whole `main` method?

Yeah this seems right to me. Callers are expected to provide their own autorelease pools, so we can probably remove the manual release here.

xixixao:
In that case #196 has the required change.

PR #200, opened 2021-01-19
Fix MTLArgumentType and MTLArgumentAccess size

### LINKED ISSUE #199: fatal assert when calling ArgumentDescriptor::set_access
I'm getting a very strange assertion with this program:

```rust
fn main() {
    let argument_descriptor = rafx::api::metal_rs::ArgumentDescriptor::new();
    println!("settings access");
    argument_descriptor.set_access(rafx::api::metal_rs::MTLArgumentAccess::ReadWrite);
    println!("has set access");

    let argument_descriptor = rafx::api::metal_rs::ArgumentDescriptor::new();
    println!("settings access");
    argument_descriptor.set_access(rafx::api::metal_rs::MTLArgumentAccess::ReadWrite);
    println!("has set access");
```

It prints

```
settings access
has set access
settings access
validateMTLArgumentAccess:27: failed assertion `access (4294967297) is not a valid MTLArgumentAccess.'
```

The bizarre thing is that I cannot reproduce this with an "empty" project. If I take the exact code and all the same dependencies and move them to another project, it works fine. The only difference I can think of at this point is the one that fails is an example of another crate and the in-theory identical project that doesn't crash links that same crate as a downstream binary.

The only thing I can think of that would cause a crash like this would be incorrect bindings or something happening before main() that causes undefined behavior. I use static variables and things like lazy_static extremely sparingly. The app runs fine if I use my vulkan backend instead of the metal backend and all other metal bindings seem to work. (The original project that is crashing calls quite a few metal APIs before reaching this point).

The only thing I saw that looks suspicious is that MTLArgumentAccess is repr(u32) and the docs point out that this is Uint, which would actually be u64 on most systems (i.e. usize)

I tried to find examples of this function actually being used but was unable to get the gfx-hal examples to call it and didn't see it used in the examples of this crate.

Just posting here hoping that someone has an idea!

kvark:
Fixes #199

### DIFF
--- a/src/argument.rs
+++ b/src/argument.rs
@@ -109,7 +109,7 @@ pub enum MTLDataType {
     RGB9E5Float = 77,
 }
 
-#[repr(u32)]
+#[repr(u64)]
 #[allow(non_camel_case_types)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLArgumentType {
@@ -121,7 +121,7 @@ pub enum MTLArgumentType {
     Imageblock = 17,
 }
 
-#[repr(u32)]
+#[repr(u64)]
 #[allow(non_camel_case_types)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLArgumentAccess {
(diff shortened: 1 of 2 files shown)


aclysma:
Made this change locally and it fixed the issue for me, thanks for looking at it so quickly! While it might be more accurate to be repr(usize) I can't think of any practical scenario where metal would be used on a 32-bit platform :)

kvark:
I think we are using `u64` consistently for `NSUInteger`, so let's not cross this bridge in this issue, but keep it in mind. Thank you for testing!

PR #201, opened 2021-01-20
Use &FenceRef instead of FenceRef in RenderCommandEncoder

aclysma:
As I understand it all Ref objects need to use &s. FYI this change is somewhat blind because my metal backend isn't bootstrapped enough to properly test it yet, so feel free to sit on it until I can :)


kvark (approved):
nice find!

kvark:
I wonder if there is a solution to limit this a compile time, something with `Pin<>` involved...

PR #206, opened 2021-03-29
Expose Drawable ID

mehmetoguzderin:
This PR exposes Drawable ID for MTLDrawable

https://developer.apple.com/documentation/metal/mtldrawable/2806860-drawableid?language=objc


kvark (approved):
Looks reasonable, thank you!

PR #208, opened 2021-04-14
Add initial HDR support

Noxime:
This allows rendering HDR content out to display by setting `wantsExtendedDynamicRangeContent`. Tested to work with gfx-hal & wgpu when using `Rgba16Float` swapchain formats. There is more work to be done when it comes to setting [EDR metadata](https://developer.apple.com/documentation/quartzcore/cametallayer/3182052-edrmetadata), but I wanted to get this small change out first.


kvark (approved):
Thank you!

PR #213, opened 2021-05-17
Add MTLBlitOption::None

chinedufn:


kvark:
Wouldn't BGRA copy work fine with MTLBlitOptionNone?

chinedufn:
Ah I didn't see that in here https://developer.apple.com/documentation/metal/mtlblitoption and didn't realize it existed 

Much better, thanks.

kvark:
The way it can be used today is via `MTLBlitOption::empty()`, and that's what we are doing in wgpu. Do we still need to add `None` constant with this in mind?

chinedufn:
Ah, when I originally opened this issue I was actually wondering how WebGPU implemented `CommandEncoder.copy_texture_to_buffer` without being able to omit the blip option. Now I see.

---

My thinking:

- If you navigate to MTLBlitOption in your IDE you only see the `blitflags!` struct definition.

- If you go to the [API docs you see a failed build](https://docs.rs/crate/metal/0.22.0). I'm not sure what docs rs' story is for mac builds these days.

- If I'm coming from a Swift/Objective-C/etc background to Rust, I'm looking for the enum variant and will be confused when I don't see it.

So, right now, if you're unfamiliar with the `bitflags` crate you have no reasonable way of knowing that it's possible to do `MTLBlitOption::empty()`.

My vote would be that `::None` is more user friendly (I would have seen it when I navigated to the source code and never opened this issue in the first place) and more consistent with the official Apple docs, with no downsides that I can think of.

So, I think that we should add it.

Of course let me know if you disagree.

kvark:
I see pros and cons about exposing None, but ultimately I'm fine either way, as long as we expose it consistently in all the bitflags.

> If I'm coming from a Swift/Objective-C/etc background to Rust, I'm looking for the enum variant and will be confused when I don't see it.

I think this is a core of the issue. This isn't an enum!
And what makes it *look* like an enum is the snake-case variants. All the bitflags in the while are all upper-cased, like `MTLBlitOption::DEPTH_FROM_DEPTH_STENCIL`.

The problem with `None` is that it's not a flag, unlike all others, and it's not necessary. One way to approach this would be to document all the bitflag types as bitflags, to avoid confusion.

Briefly looking at the other types, I see `MTLFunctionOptions` which already has `None` variant. So at least adding None here is somewhat consistent, hence we can proceed.

In the longer term, I'd like to see real idiomatic use of bitflags here, but it's a minor wish :).

chinedufn:
> I think this is a core of the issue. This isn't an enum!

Ahhhh, right good point thanks.

PR #214, opened 2021-06-16
Replace cocoa-foundation dependency to core-graphics-types

kvark:
`core-graphics-types` is a smaller dependency, which I think we can accept to depend on across the stack.
`cocoa-foundation` doesn't appear to be necessary.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1706,7 +1706,6 @@ impl DeviceRef {
             let library: *mut MTLLibrary = msg_send![self, newLibraryWithSource:source
                                                                         options:options
                                                                           error:&mut err];
-            let () = msg_send![source, release];
             if !err.is_null() {
                 let desc: *mut Object = msg_send![err, localizedDescription];
                 let compile_error: *const c_char = msg_send![desc, UTF8String];
@@ -1730,9 +1729,6 @@ impl DeviceRef {
                 msg_send![self, newLibraryWithFile:filename.as_ref()
                                              error:&mut err]
             };
-
-            let () = msg_send![filename, release];
-
             Ok(Library::from_ptr(library))
         }
     }
--- a/src/device.rs
+++ b/src/device.rs
@@ -11,12 +11,7 @@ use block::{Block, ConcreteBlock};
 use foreign_types::ForeignType;
 use objc::runtime::{Object, NO, YES};
 
-use std::{
-    ffi::CStr,
-    os::raw::c_char,
-    path::Path,
-    ptr,
-};
+use std::{ffi::CStr, os::raw::c_char, path::Path, ptr};
 
 #[allow(non_camel_case_types)]
 #[repr(u64)]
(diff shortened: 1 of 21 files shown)


src/lib.rs
```diff
@@ -17,33 +17,26 @@ extern crate objc;
...
 use foreign_types::ForeignType;
 use objc::runtime::{Object, NO, YES};
 
 #[cfg(target_pointer_width = "64")]
-pub type CGFloat = f64;
+pub type NSInteger = i64;
 #[cfg(not(target_pointer_width = "64"))]
-pub type CGFloat = f32;
-
-#[repr(C)]
```
chinedufn, line 35: nice

src/lib.rs
```diff
@@ -17,33 +17,26 @@ extern crate objc;
...
-impl CGSize {
-    pub fn new(width: f64, height: f64) -> Self {
-        CGSize { width, height }
-    }
-}
+pub type NSInteger = i32;
+#[cfg(target_pointer_width = "64")]
+pub type NSUInteger = u64;
+#[cfg(target_pointer_width = "32")]
+pub type NSUInteger = u32;
```
madsmtm, line 39: I would probably prefer:
```rust
pub type NSInteger = isize;
pub type NSUInteger = usize;
```

madsmtm (approved):

src/lib.rs
kvark: that's what I started with, but the problem here is that `NSUInteger` is not actually a `usize`. So whoever uses the real `NSUInteger` will get into a problem: Rust will yell at them since `usize != u64`.

madsmtm: By "real `NSUInteger`", do you mean `cocoa-foundation`'s?

kvark: Yes, that one. If somebody depends on cocoa-foundation, we still want to be compatible with them

src/device.rs
```diff
@@ -1705,16 +1700,12 @@ impl DeviceRef {
         src: &str,
         options: &CompileOptionsRef,
     ) -> Result<Library, String> {
-        use cocoa_foundation::base::nil as cocoa_nil;
-        use cocoa_foundation::foundation::NSString as cocoa_NSString;
-
+        let source = nsstring_from_str(src);
```
grovesNL, line 1703: Is it alright that we're relying on `autorelease` for these now? I guess we probably already wrap these with an autorelease pool when we use them in gfx/wgpu

grovesNL (approved):
Thanks!

src/device.rs
kvark: I got bitten by this since I wrote the PR.
So yeah, you rightfully noted that we are now relying on ARP to clean the string up.
I don't think gfx or wgpu backends currently wrap the pipeline creation in ARP, but they could. So technically, it's going to be leaking the strings with pipeline creation. I think it's fine, we can live with this (and address this in downstream if we want).

PR #216, opened 2021-06-17
make CGFloat and CGSize public again

ccbrown:
In https://github.com/gfx-rs/metal-rs/pull/214, `CGFloat` and `CGSize` went from being public to private, which breaks compatibility with the gfx backend. This makes them public again.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,7 @@ use std::{
     os::raw::c_void,
 };
 
-use core_graphics_types::{base::CGFloat, geometry::CGSize};
+pub use core_graphics_types::{base::CGFloat, geometry::CGSize};
 use foreign_types::ForeignType;
 use objc::runtime::{Object, NO, YES};
 


kvark:
I don't think we need to re-export these types. The users of metal-rs should just depend on core-graphics-types themselves.

ccbrown:
That's fair but now it's no longer possible to use newer metal-rs changes with gfx via Cargo patch directives.

I would put in a PR to use core-graphics-types directly in gfx, but unless I use an unreleased metal-rs revision, I don't believe that would work either as the types would have different definitions.

I'm fine with closing this, but it is unfortunate that despite your incredible responsiveness, I'll have to use a fork until the next release. 😕

kvark:
> it's no longer possible to use newer metal-rs changes with gfx via Cargo patch directives.

Generally speaking, metal-rs on master is not guaranteed to be compatible with last released metal-rs. So patching it like this is not an encouraged workflow. Instead, we can land patches to v0.22 branch, and it can be used for your cargo patch. These patches would have to be backwards compatible.

> I would put in a PR to use core-graphics-types directly in gfx, but unless I use an unreleased metal-rs revision, I don't believe that would work either as the types would have different definitions.

It's fine for gfx-rs to use metal-rs from github, up until the point where gfx-rs needs to be released. Speaking of which... we were just planning on doing exactly this with @cwfitzgerald ! So I'm just going to go ahead and publish metal-0.23. If you want to help by updating gfx-rs accordingly, this would be much welcome!

ccbrown:
Excellent, thank you! I'll likely update gfx then when I next get the chance.

PR #217, opened 2021-06-30
Add MPSMatrix and related classes

yumios:
Add some classes to be able to calculate matrix multiplication on metal. And add an example that executes matrix multiplication commands.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,10 @@ name = "mps"
 required-features = ["mps"]
 
 [[example]]
+name = "mps-matrix"
+required-features = ["mps"]
+
+[[example]]
 name = "embedded-lib"
 path = "examples/compute/embedded-lib.rs"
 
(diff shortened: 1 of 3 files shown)


kvark:
This is a part of the Metal performance shaders, right? What volume of the API surface are we considering here?
I think we discussed it briefly in the past, and I suggested to have the MPS in a separate library that depends on metal-rs, like metal-perf-shaders or something. But I haven't really used it, so open to opinions!

yumios:
Sorry for not knowing that you already suggested having the MPS in a separate library.
I'm going to add various MPS features about MPSCNN and MPSMatrix to use in my applications. So, I agree with your idea because MPS looks too huge to add to metal-rs and I feel it isn't often used by existing metal-rs users.
I'll close this PR and write a new library for binding MPS API. Thank you for your feedback! It helped me a lot.

kvark:
Thank you! Looking forward to see this emerging:)

PR #221, opened 2021-11-22
Add support for getting/setting colorspace, add missing GPU families

aclysma:
Is adding core-graphics as a dependency ok? Is there an alternative?


kvark:
Let's see if anybody responds to https://github.com/servo/core-foundation-rs/issues/479

PR #223, opened 2021-12-02
Add new_texture_with_io_surface_plane

xiaopengli89:


kvark:
It's unfortunate to drag the iosurface dependency in. It breaks often and depends on a whole bunch of stuff - https://crates.io/crates/io-surface/0.15.1/dependencies

kvark:
Filed https://github.com/servo/core-foundation-rs/issues/483 for a better solution

xiaopengli89:
All right

PR #227, opened 2022-02-06
Fix incorrect MTLRenderStages constants

aclysma:
This potentially fixes synchronization as these constants are passed to APIs like `waitForFence(_ fence: MTLFence, 
           before stages: MTLRenderStages)`

https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1648378-waitforfence

Would appreciate if someone could properly double-check this. I can see from the metal frame capture that this fixed incorrect values for me, and these docs show the constants being used in other bindings. https://docs.microsoft.com/en-us/dotnet/api/metal.mtlrenderstages?view=xamarin-ios-sdk-12 I was not able to find the correct values published by apple. (Maybe their C++ bindings would have them)

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -139,8 +139,8 @@ impl FenceRef {
 #[repr(u64)]
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum MTLRenderStages {
-    Vertex = 0,
-    Fragment = 1,
+    Vertex = 1,
+    Fragment = 2,
 }
 
 const BLOCK_HAS_COPY_DISPOSE: i32 = 0x02000000;


aclysma:
Another source https://github.com/bkaradzic/metal-cpp/blob/0c40582f703eadfe507c8a147f3d2e7b9aca8d5c/Metal/MTLRenderCommandEncoder.hpp#L140

(Maybe it's worth adding an entry for Tile)

grovesNL:
You can see the values on the individual constant like https://developer.apple.com/documentation/metal/mtlrenderstages/mtlrenderstagevertex?language=objc, so this change seems correct

grovesNL (approved):
Thank you!

PR #228, opened 2022-02-06
Add new use resources methods

mxpv:
This PR:
- Changes `MTLRenderStages` to bitflags as there are cases when we want to combine stages (below).
- Current `useResource` method is [deprecated](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/2866168-useresource). The [new one](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/3043404-useresource?language=objc) includes `stages` param.
- Adds `useResources` and `useHeaps` to pass slices of resources.
- Updates `argument-buffer` example to demonstrate how to use `use_resources`.
- Adds new `bindless` example that demonstrates how to bind lots of textures all at once with argument buffers.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -674,15 +674,88 @@ impl RenderCommandEncoderRef {
     // fn setVertexBuffers_offsets_withRange(self, buffers: *const id, offsets: *const NSUInteger, range: NSRange);
     // fn setVertexSamplerStates_lodMinClamps_lodMaxClamps_withRange(self, samplers: *const id, lodMinClamps: *const f32, lodMaxClamps: *const f32, range: NSRange);
 
-    pub fn use_resource(&self, resource: &ResourceRef, usage: MTLResourceUsage) {
+    /// Adds an untracked resource to the render pass, specifying which render stages need it.
+    ///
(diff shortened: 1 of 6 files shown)


src/encoder.rs
```diff
@@ -674,15 +674,88 @@ impl RenderCommandEncoderRef {
...
-    pub fn use_resource(&self, resource: &ResourceRef, usage: MTLResourceUsage) {
+    /// Adds an untracked resource to the render pass, specifying which render stages need it.
+    ///
+    /// Availability: iOS 13.0+, macOS 10.15+
+    ///
+    /// # Arguments
+    /// * `resource`: A resource within an argument buffer.
+    /// * `usage`: Options for describing how a graphics function uses the resource.
+    /// * `stages`: The render stages where the resource must be resident.
+    pub fn use_resource(
```
kvark, line 685: This doesn't seem right. First, the old method is still useful if we aren't on `10.15`.
Secondly, they are named differently in Metal docs: one is `useResource`, the other is `use`.

I admit we can't call the new one `use`. Maybe we can call it something like `useResourceAt` or something similar?

mxpv: Updated

kvark (approved):

Issue #5, opened 2017-07-12
Host documentation


kvark:
Most of the project rely on `docs.rs` for documentation, however it only builds Linux at the moment. We need to find a good way of providing the documentation here.

cc @fkaa @JohnColanduoni

kvark:
Related - https://github.com/onur/docs.rs/pull/73

mjadczak:
Also related to this: https://github.com/onur/docs.rs/issues/93

IsaccBarker:
I know this is a really old issue, but I found a way to do it with GitHub's new Actions to Pages functionality. Below is an example, and you can find [working documentation hosted on my fork.](https://github.com/IsaccBarker/metal-rs):

```yaml
name: Rust

on:
  push:
    branches: [ "master" ]
  pull_request:
    branches: [ "master" ]
  workflow_dispatch:

permissions:
  contents: read
  pages: write
  id-token: write

concurrency:
  group: "pages"
  cancel-in-progress: true

env:
  CARGO_TERM_COLOR: always

jobs:
  build:
    runs-on: macos-latest
    steps:
    - uses: actions/checkout@v3
    - name: Document
      run: cargo doc --verbose
#    - name: Packaging
#      run: tar -zcvf github-pages.tar.gz ./target/doc
    - name: Packaging
      run: mv target/doc _site
    - name: Upload artifact
      uses: actions/upload-pages-artifact@v1
  deploy:
    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}
    runs-on: ubuntu-latest
    needs: build
    steps:
      - name: Deploy to GitHub Pages
        id: deployment
        uses: actions/deploy-pages@v1
```
You have to set it up in the repository settings though.

notgull:
> Most of the project rely on docs.rs for documentation, however it only builds Linux at the moment. 

Since this is no longer true, should this issue be closed?

IsaccBarker:
If this still true? The [about builds page for docs.rs](https://docs.rs/about/builds) says:
> All targets other than x86_64-unknown-linux-gnu are cross-compiled. For implementation reasons, this is unlikely to change for the foreseeable future. 

Correct me if I'm wrong, but wouldn't out-of-ecosystem platform specific libraries still be needed, that couldn't be brought in?

kvark:
@IsaccBarker would you mind making a PR?
What do I need to set up?

madsmtm:
[...]

Issue #225, opened 2021-12-19
Make auto-release pools visible


kvark:
Inspired by #224
All methods that return auto-referenced objects have this common property: they assume ARP alive, and they could return a reference with a lifetime not greater than that ARP. This can be directly expressed with the type system instead of being implicit:
```
let bar = {
  let arp = metal::AutoReleasePool::new();
  let foo = device.create_foo(&arp);
  foo.to_owned()
};
```

Notice how ARP becomes a regular object with RAII instead of being closure-based.
Why this is cool?
1. ARP is explicitly used, therefore no need to fight Rust "unused variable" warnings that often occur with similar RAII APIs
2. metal-rs clearly expresses the dependency on ARP, there is no assumption taking place
3. everything is forcefully safe, since returned references live no longer than the ARP
4. metal-rs is zero overhead, since it doesn't do "retain" just to be able to return an owned object

madsmtm:
Also partly discussed in https://github.com/gfx-rs/metal-rs/issues/222.

Quick notes: The following two examples would be UB:
```rust
let bar = {
  let outer = metal::AutoReleasePool::new();
  let foo = {
    let inner = metal::AutoReleasePool::new();
    // Released with inner pool, but bound to the outer pool
    device.create_foo(&outer)
  };
  foo.to_owned()
};
```

```rust
let a = metal::AutoReleasePool::new();
let b = metal::AutoReleasePool::new();
// Drop order is required to be the inverse of creation (drop b first, then a)
drop(a);
drop(b);
```

So I don't think you could make a safe (zero-cost) API using just RAII, although your proposal would be substantially better than the status-quo!

I've been working on this idea as well, see related issues in `objc`: https://github.com/SSheldon/rust-objc/pull/103 and https://github.com/SSheldon/rust-objc/issues/95.

Another reference is in my fork, the [current implementation](https://github.com/madsmtm/objc2/blob/fcecbf0b5700e3c0b21be7827b2086e9131ac987/objc2/src/rc/autorelease.rs) (uses closures + thread locals in debug mode and `auto` traits on `nightly`) and [an example usage](https://github.com/madsmtm/objc2/blob/master/objc2-foundation/src/string.rs#L35-L87) (complicated by `-[NSString UTF8String]` being weirder than most).

Issue #222, opened 2021-11-23
Investigate use of objc_retainAutoreleasedReturnValue


raphlinus:
Autorelease is a major pain point for Apple integration, and users of this library in particular. It forces an uncomfortable tradeoff between performance and safety, and is quite prone to memory leaks if the calling application doesn't scope the autorelease pools properly.

Using Apple-controlled compilers (ie the official Swift bindings), there's a good solution in the form of [objc_retainAutoreleasedReturnValue]. However, there are challenges for us to use it.

* It's not currently present in the objc bindings.

* It depends on some very specific magic to work, notably [bit patterns](https://www.galloway.me.uk/2012/02/how-does-objc_retainautoreleasedreturnvalue-work/) in the generated asm for the call to the allocating method. It's not obvious to me the best way to make this robust; it might require upstream support in Rust or possibly some other mechanism to generate the proper asm.

* The [objc infrastructure may be undergoing a split](https://github.com/SSheldon/rust-objc/issues/101), making it less clear where this would go.

This issue came up in the recent [Hacker News discussion of the official C++ Metal bindings]. Those bindings also don't have a great story around autorelease (I was hopeful they did and we'd be able to adapt it). See [this thread](https://news.ycombinator.com/item?id=29311736) in particular for the autorelease-specific discussion, including very helpful info from @comex.

[objc_retainAutoreleasedReturnValue]: https://clang.llvm.org/docs/AutomaticReferenceCounting.html#arc-runtime-objc-retainautoreleasedreturnvalue

[Hacker News discussion of the official C++ Metal bindings]: https://news.ycombinator.com/item?id=29289761

drewcrawford:
I have looked into this problem for my crates at some length. The short version is that x64 is “easy enough” if you are willing to very carefully control inlining, etc. But it does require incredibly close coordination between a metal binding and an objc crate (and everything in between - blocks is another common example for block-based APIs) as you need one to directly follow the other in the compiled binary, accounting for optimizations etc. So in terms of “where this would go” the answer is unfortunately “yes”. 

arm/etc has the additional complication of requiring particular instruction patterns. Stabilizing asm! would help here.  the only solution I know of for stable rust is to emit thunks in an assembler, which introduces additional stack frames and spoils many other optimizations so is pretty undesirable. 

In my workloads this is usually not the heaviest rock optimization-wise but once you do lift the heavier rocks it does start to matter.

raphlinus:
[...]

Issue #218, opened 2021-10-19
Don't use auto-release pool in string-related methods


kvark:
We have `set_label` methods, which use `nsstring_from_str`, which uses auto-release pool. This is inviting for leaks in user code. There is no reason why `set_label` should use any auto-release pools, technically.

Proposal - ~~split `nsstring_as_str` into separate methods for creation of an `NSString` and borrowing it as `str`. This way, the calling code is responsible for freeing it when needed.~~ don't autorelease `NSString`. Instead, return an object that the caller can free explicitly by dropping. The caller is our own method in this case, so we can ensure the lifetime is correct.

PR #241, opened 2022-09-01
Use `objc2` instead of `objc`

madsmtm:
Replace `objc` with my fork, [`objc2`](https://github.com/madsmtm/objc2).

This crate contains many improvements, in particular:
- Better tools for protecting against UB (see first commit)
- You no longer have to worry about using `bool` in `msg_send!`.
- The `msg_send_id!` macro greatly helps with follow memory management rules (similar to Objective-C ARC), potentially also alleviating a lot of the need for autoreleasepools.
- The `extern_protocol!` macro is used in place of `foreign_obj_type!`

Notable breaking changes:
- `objc2::Message` differs from `objc::Message`, so message sending is not compatible
- `NSUInteger` is now `usize` instead of `u32`/`u64` depending on platform (equivalently for `NSInteger`)
- There is no longer three types for each class (e.g. `MTLFoo`, `Foo` and `FooRef`); these have been combined into one: `Foo`. Pointers to this can be used across FFI boundaries, and `objc2::rc::Id<Foo, Shared>` can be used as a strong pointer. In table form:

  | Previously | Now |
  | --- | --- |
  | `*mut MTLFoo` | `*mut Foo` |
  | `&FooRef` | `&Foo` |
  | `Foo` | `Id<Foo, Shared>` |

When done, this should resolve the following issues:
- Fixes https://github.com/gfx-rs/metal-rs/issues/5 now
- Later: https://github.com/gfx-rs/metal-rs/issues/222
- Later: https://github.com/gfx-rs/metal-rs/issues/218
- Later: https://github.com/gfx-rs/metal-rs/issues/128
- Later: https://github.com/gfx-rs/metal-rs/issues/86
- Later: https://github.com/gfx-rs/metal-rs/issues/99
- Later: https://github.com/gfx-rs/metal-rs/issues/225
- Later: https://github.com/gfx-rs/metal-rs/issues/58
- Later: https://github.com/gfx-rs/metal-rs/issues/282
- Later: https://github.com/gfx-rs/metal-rs/issues/284

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,46 +28,10 @@ use std::{
     os::raw::c_void,
 };
 
-use core_graphics_types::{base::CGFloat, geometry::CGSize};
 use foreign_types::ForeignType;
 pub(crate) use objc2::encode::{Encode, Encoding, RefEncode};
+pub use objc2::foundation::{CGFloat, NSInteger, NSRange, NSSize as CGSize, NSUInteger};
 use objc2::runtime::{Bool, Object, Protocol};
-
-/// See <https://developer.apple.com/documentation/objectivec/nsinteger>
-#[cfg(target_pointer_width = "64")]
-pub type NSInteger = i64;
-
-/// See <https://developer.apple.com/documentation/objectivec/nsinteger>
-#[cfg(not(target_pointer_width = "64"))]
-pub type NSInteger = i32;
-
-/// See <https://developer.apple.com/documentation/objectivec/nsuinteger>
-#[cfg(target_pointer_width = "64")]
-pub type NSUInteger = u64;
-
-/// See <https://developer.apple.com/documentation/objectivec/nsuinteger>
-#[cfg(target_pointer_width = "32")]
-pub type NSUInteger = u32;
-
-/// See <https://developer.apple.com/documentation/foundation/nsrange>
-#[repr(C)]
-#[derive(Copy, Clone)]
-pub struct NSRange {
-    pub location: NSUInteger,
-    pub length: NSUInteger,
-}
-
-unsafe impl objc2::Encode for NSRange {
-    const ENCODING: objc2::Encoding =
-        objc2::Encoding::Struct("_NSRange", &[NSUInteger::ENCODING, NSUInteger::ENCODING]);
-}
-
-impl NSRange {
-    #[inline]
-    pub fn new(location: NSUInteger, length: NSUInteger) -> NSRange {
-        NSRange { location, length }
-    }
-}
 
 fn nsstring_as_str(nsstr: &Object) -> &str {
     let bytes = unsafe {
@@ -76,7 +40,7 @@ fn nsstring_as_str(nsstr: &Object) -> &str {
     };
     let len: NSUInteger = unsafe { msg_send![nsstr, length] };
     unsafe {
-        let bytes = std::slice::from_raw_parts(bytes, len as usize);
+        let bytes = std::slice::from_raw_parts(bytes, len);
         std::str::from_utf8(bytes).unwrap()
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,7 +14,7 @@ pub extern crate bitflags;
 #[macro_use]
 pub extern crate log;
 #[macro_use]
-pub extern crate objc;
+pub extern crate objc2;
 #[macro_use]
 pub extern crate foreign_types;
 #[macro_use]
@@ -30,7 +30,8 @@ use std::{
 
 use core_graphics_types::{base::CGFloat, geometry::CGSize};
 use foreign_types::ForeignType;
-use objc::runtime::{Object, NO, YES};
+pub(crate) use objc2::encode::{Encode, Encoding, RefEncode};
+use objc2::runtime::{Bool, Object, Protocol};
 
 /// See <https://developer.apple.com/documentation/objectivec/nsinteger>
 #[cfg(target_pointer_width = "64")]
@@ -56,6 +57,11 @@ pub struct NSRange {
     pub length: NSUInteger,
 }
 
+unsafe impl objc2::Encode for NSRange {
+    const ENCODING: objc2::Encoding =
+        objc2::Encoding::Struct("_NSRange", &[NSUInteger::ENCODING, NSUInteger::ENCODING]);
+}
+
 impl NSRange {
     #[inline]
     pub fn new(location: NSUInteger, length: NSUInteger) -> NSRange {
@@ -63,7 +69,7 @@ impl NSRange {
     }
 }
 
-fn nsstring_as_str(nsstr: &objc::runtime::Object) -> &str {
+fn nsstring_as_str(nsstr: &Object) -> &str {
     let bytes = unsafe {
         let bytes: *const std::os::raw::c_char = msg_send![nsstr, UTF8String];
         bytes as *const u8
@@ -75,20 +81,20 @@ fn nsstring_as_str(nsstr: &objc::runtime::Object) -> &str {
     }
 }
 
-fn nsstring_from_str(string: &str) -> *mut objc::runtime::Object {
+fn nsstring_from_str(string: &str) -> *mut Object {
     const UTF8_ENCODING: usize = 4;
 
     let cls = class!(NSString);
     let bytes = string.as_ptr() as *const c_void;
     unsafe {
-        let obj: *mut objc::runtime::Object = msg_send![cls, alloc];
-        let obj: *mut objc::runtime::Object = msg_send![
+        let obj: *mut Object = msg_send![cls, alloc];
+        let obj: *mut Object = msg_send![
             obj,
             initWithBytes:bytes
             length:string.len()
             encoding:UTF8_ENCODING
         ];
-        let _: *mut c_void = msg_send![obj, autorelease];
+        let _: *mut Object = msg_send![obj, autorelease];
         obj
     }
 }
@@ -174,15 +180,24 @@ macro_rules! foreign_obj_type {
             }
         }
 
-        unsafe impl ::objc::Message for $raw_ident {
+        unsafe impl ::objc2::encode::Encode for $owned_ident {
+            const ENCODING: ::objc2::Encoding = ::objc2::Encoding::Object;
         }
-        unsafe impl ::objc::Message for paste!{[<$owned_ident Ref>]} {
+        unsafe impl ::objc2::encode::RefEncode for $raw_ident {
+            const ENCODING_REF: ::objc2::Encoding = ::objc2::Encoding::Object;
+        }
+        unsafe impl ::objc2::encode::RefEncode for paste!{[<$owned_ident Ref>]} {
+            const ENCODING_REF: ::objc2::Encoding = ::objc2::Encoding::Object;
+        }
+        unsafe impl ::objc2::Message for $raw_ident {
+        }
+        unsafe impl ::objc2::Message for paste!{[<$owned_ident Ref>]} {
         }
 
         impl ::std::fmt::Debug for paste!{[<$owned_ident Ref>]} {
             fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
                 unsafe {
-                    let string: *mut ::objc::runtime::Object = msg_send![self, debugDescription];
+                    let string: *mut ::objc2::runtime::Object = msg_send![self, debugDescription];
                     write!(f, "{}", crate::nsstring_as_str(&*string))
                 }
             }
@@ -201,7 +216,7 @@ macro_rules! try_objc {
         $err_name: ident => $body:expr
     } => {
         {
-            let mut $err_name: *mut Object = ::std::ptr::null_mut();
+            let mut $err_name: *mut ::objc2::runtime::Object = ::std::ptr::null_mut();
             let value = $body;
             if !$err_name.is_null() {
                 let desc: *mut Object = msg_send![$err_name, localizedDescription];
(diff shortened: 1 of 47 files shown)


src/sync.rs
```diff
@@ -81,58 +89,53 @@ impl SharedEventRef {
...
     pub fn device(&self) -> &DeviceRef {
         unsafe { msg_send![self, device] }
     }
 
-    pub fn label(&self) -> &str {
-        unsafe {
-            let label = msg_send![self, label];
-            crate::nsstring_as_str(label)
-        }
+    pub fn label(&self) -> Id<NSString, Shared> {
```
madsmtm, line 132: Note that we returned `&str` before, but that is not sound since the string could be released by an autoreleasepool; instead we now use `Id<NSString, Shared>`, which should be faster because of the internal usage of `objc_retainAutoreleasedReturnValue`, and it allows the user to choose which conversion they'd like to do (e.g. `.label().to_string()` or `.label().as_str(pool)`).

src/sync.rs
```diff
@@ -153,6 +156,10 @@ bitflags! {
     }
 }
 
+unsafe impl Encode for MTLRenderStages {
+    const ENCODING: Encoding = u64::ENCODING;
+}
```
madsmtm, line 161: These `Encode` implementations help us catch bugs when the `verify_message` feature is enabled, and in the future [when debug assertions are enabled](https://github.com/madsmtm/objc2/pull/198); this is a _huge_ boost for soundness!

madsmtm (commented):
@kvark @cwfitzgerald I think I'm ready for some feedback on this before I proceed; there's a lot of things going on here, but I'd like you to take a look at e.g. [the last commit](https://github.com/gfx-rs/metal-rs/pull/241/commits/9d20a2ffa0ef7bb697765da7833384ed44ea4bf5), where I've changed things to use the `extern_class!`, `extern_protocol!` and `msg_send_id!` macros.

Feel free to ask me to clarify things about `objc2` in general, I recognize that the documentation of it is not (yet) ideal.

i509VCB:
The pull request title appears to be wrong, shouldn't objc and objc2 be swapped in the title?

madsmtm:
An update on this:

I decided to make an automatically generated wrapper for Objective-C frameworks called [`icrate`](https://docs.rs/icrate/latest/icrate/). There are a few downsides to an automatically generated library, but most of these can be mitigated by manual additions, which I've explicitly designed `icrate` to support.

In https://github.com/madsmtm/objc2/pull/329 I've added support for Metal, with the primary addition being the file [`crates/header-translator/src/data/Metal.rs`](https://github.com/madsmtm/objc2/pull/329/files#diff-a70d31be47d3a327a74e035b955babd53ba188082dc2d3275107040687514378), which details which methods are safe (methods are naturally `unsafe` by default).

This library fixes the same issues as described in the PR description, as well as paving a way forwards for many other improvements, such as:
- Automatic [API availability](https://github.com/madsmtm/objc2/issues/266) / deprecation (so that users only call new methods when available).
- Quick SDK updating (e.g. https://github.com/gfx-rs/metal-rs/pull/252 would've just been: Download the new SDK, rerun the build script, and mark safe methods).
- Reduced risk of mistakes that lead to unsoundness (for example, _all_ of `metal`'s enum definitions are unsound, since a future version of the OS is allowed to add more variants, and that would break Rust's assumption that the enum is exhaustive).

I'm still working on improving how protocols are mapped, which has a big effect on Metal - would like some help with this in https://github.com/madsmtm/objc2/issues/291 (and maybe Metal protocols should work differently from other frameworks?).

---

I would also like to discuss with the maintainers (@kvark @cwfitzgerald); what do you think the best way forwards is?

I shall be blunt 😉: My ideal future would be for the `metal` crate just being a `pub use icrate::Metal::*`.

To achieve this I'm very open to changing things that you may be unhappy with, like the [method naming scheme](https://github.com/madsmtm/objc2/issues/284) or adding collaborators to `icrate`/`objc2` if you would be more comfortable with it if the bus factor was decreased?

I'm reachable on Matrix as `@madsmtm:matrix.org`, and will probably see a ping on `#wgpu:matrix.org` as well, if you want to have a more synchronous discussion.

grovesNL:
> I would also like to discuss with the maintainers (@kvark @cwfitzgerald); what do you think the best way forwards is?
>
> I shall be blunt 😉: My ideal future would be for the metal crate just being a pub use icrate::Metal::*.
>
> To achieve this I'm very open to changing things that you may be unhappy with, like the https://github.com/madsmtm/objc2/issues/284 or adding collaborators to icrate/objc2 if you would be more comfortable with it if the bus factor was decreased?

Looks really great so far! Awesome work and I'm excited about fixing some of the autorelease pool and selector issues we've discussed over the years.

I'm a little hesitant about generating all bindings and placing them all in icrate. A few other projects have taken a similar approach to replace objc (e.g., objr, fruity, objrs, etc.) but have never achieved broad adoption by the existing ecosystem, so I'm not confident that objc2/icrate will become the standard yet (though I acknowledge that this is a bit of a chicken-and-egg problem).

For example, ideally I don't think it should be the metal's crate responsibility to say that we only work with `NSColor`s that originate from objc2/icrate, but I could see us easily stuck in that position if we switch.

Any of the following would help a lot:
- more collaborators to reduce the bus factor
- confidence that the generated bindings are a net time saver/reduce maintenance complexity
  - e.g., lowering the amount of manual work instead of often having to patch the generator itself or write a lot of manual fixes on top of the output
- clear signs that the ecosystem will really move in this direction
  - e.g., maybe upgrading the existing ecosystem instead of replacing it where possible
- better understanding around increases or decreases in compilation time or runtime performance

I'm also curious if there's a way we could adopt generated bindings internally without breaking the existing API too much (instead of, or as a temporary measure while we work towards being able to do `pub use icrate::Metal::*`).

> Automatic API availability / deprecation (so that users only call new methods when available).

For what it's worth, we wouldn't want compile-time detection for this in all cases. For example, wgpu's Metal backend would want to be able to compile against the highest possible deployment target and decide at runtime which methods to use.

madsmtm:
> Looks really great so far! Awesome work and I'm excited about fixing some of the autorelease pool and selector issues we've discussed over the years.

Thanks! It's great to hear you raise some concerns, I'll try to respond to them below:

> I'm a little hesitant about generating all bindings and placing them all in icrate. A few other projects have taken a similar approach to replace objc (e.g., objr, fruity, objrs, etc.) but have never achieved broad adoption by the existing ecosystem, so I'm not confident that objc2/icrate will become the standard yet (though I acknowledge that this is a bit of a chicken-and-egg problem).

Yeah, I'm aware, though still very much a valid point. If it helps, I'm the maintainer of `winit` and I plan to use it there, which will likely cause pressure on `wgpu` to at the very least switch to `objc2` to reduce the dependency tree.

> For example, ideally I don't think it should be the metal's crate responsibility to say that we only work with `NSColor`s that originate from objc2/icrate, but I could see us easily stuck in that position if we switch.

Metal doesn't actually use `NSColor`, but I get what you mean with `NSString`, `NSArray`, `NSDictionary`, and so on ;).

In my defence, `metal` is already tied to `objc`'s versioning, though yeah the problem does get worse with a large crate (and probably for a while a fast-moving target) like `icrate`.

> Any of the following would help a lot:
>
>     * more collaborators to reduce the bus factor

Yeah, will hopefully get better with time. I'll probably make a call out on TWiR at some point for starters.

>     * confidence that the generated bindings are a net time saver/reduce maintenance complexity
>
>       * e.g., lowering the amount of manual work instead of often having to patch the generator itself or write a lot of manual fixes on top of the output

I think it actually does really well on this mark already (I was surprised myself, I thought I'd have to implement a bunch more hacks for Metal), the only real burden was in creating the data file for which methods are `unsafe`, which is _as it should be_, that part _should_ require attention.

>     * clear signs that the ecosystem will really move in this direction

That's indeed on my roadmap: https://github.com/madsmtm/objc2/issues/174

>       * e.g., maybe upgrading the existing ecosystem instead of replacing it where possible

Could you provide an example of what you mean here?

>     * better understanding around increases or decreases in compilation time or runtime performance

Wrt. compilation time, I'm working on a system to help with this in https://github.com/madsmtm/objc2/issues/311.

Runtime performance is kinda difficult to prove before the entirety of the migration is done, but I can at least say that the project has [assembly-level tests](https://github.com/madsmtm/objc2/tree/59c6f44a5dd7a5d495baaea709841d5fb750f7e8/crates/test-assembly) for ensuring that the autorelease optimizations _do_ happen.

> I'm also curious if there's a way we could adopt generated bindings internally without breaking the existing API too much (instead of, or as a temporary measure while we work towards being able to do `pub use icrate::Metal::*`).

It is very much possible to do something like this:

```rust
use icrate::Metal::MTLDepthStencilDescriptor;

#[repr(u64)]
#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
pub enum MTLCompareFunction {
    Never = 0,
    Less = 1,
    Equal = 2,
    LessEqual = 3,
    Greater = 4,
    NotEqual = 5,
    GreaterEqual = 6,
    Always = 7,
}

// from_icrate and into_icrate for MTLCompareFunction, unless we simply export icrate's MTLCompareFunction


foreign_obj_type! {
    type CType = MTLDepthStencilDescriptor;
    pub struct DepthStencilDescriptor;
}
// Will now translate to roughly
{
    pub struct DepthStencilDescriptor(Id<MTLDepthStencilDescriptor, Shared>);

    #[repr(transparent)]
    pub struct DepthStencilDescriptorRef(MTLDepthStencilDescriptor);

    impl Deref<Target = DepthStencilDescriptorRef> for DepthStencilDescriptor { ... }

    impl DepthStencilDescriptor {
        // Could also be a `From` implementation, depending on how much you want
        // to expose that you're using `icrate` under the hood.
        fn from_icrate(obj: Id<MTLDepthStencilDescriptor, Shared>) -> Self {
            Self(obj)
        }
    
        fn into_icrate(...) { ... }
    }
    
    // from_icrate and into_icrate for DepthStencilDescriptorRef
}


impl DepthStencilDescriptor {
    pub fn new() -> Self {
        Self(MTLDepthStencilDescriptor::new())
    }
}

impl DepthStencilDescriptorRef {
    pub fn depth_compare_function(&self) -> MTLCompareFunction {
        MTLCompareFunction::from_objc2(self.0.depthCompareFunction())
    }

    pub fn set_depth_compare_function(&self, func: MTLCompareFunction) {
        self.0.setDepthCompareFunction(func.into_objc2());
    }

    pub fn depth_write_enabled(&self) -> bool {
        self.0.isDepthWriteEnabled();
    }

    pub fn set_depth_write_enabled(&self, enabled: bool) {
        self.0.setDepthWriteEnabled(enabled);
    }

    pub fn front_face_stencil(&self) -> Option<StencilDescriptor> {
        self.frontFaceStencil().map(StencilDescriptor::from_objc2)
    }

    pub fn set_front_face_stencil(&self, descriptor: Option<&StencilDescriptorRef>) {
        self.setFrontFaceStencil(descriptor.map(|d| d.into_objc2()))
    }

    // ...

    // Inefficient, but correct! An alternative would be to return a newtype `MetalString(Id<NSString, Shared>)` instead.
    pub fn label(&self) -> String {
        self.0.label().to_string()
    }

    pub fn set_label(&self, label: &str) {
        self.0.setLabel(&NSString::from_str(label));
    }
}
```

And this `-sys` pattern is what I would usually do for normal C-APIs; however, Objective-C is different in that they have quite high type-safety, which means, as you can see, a lot of things just end up being duplicated work.

But yes, it would definitely be a way forward, and I could probably do it at some point if you want.

> > Automatic API availability / deprecation (so that users only call new methods when available).
> 
> For what it's worth, we wouldn't want compile-time detection for this in all cases. For example, wgpu's Metal backend would want to be able to compile against the highest possible deployment target and decide at runtime which methods to use.

The plan is to have the checks be made at runtime, but in a way such that you'll get a panic if you use some functionality _without_ having made the runtime check beforehand, similar to Swift's `#available` (which, yeah, will require a bit of trickery to get to work in Rust).

---

Again, thanks for the kind response!

madsmtm:
Okay, so to get this moving forwards, I've reduced this PR to only a "move to `objc2`", and have kept the more fancy features and improvements out for now - they will come later, but I'd like to get the ball rolling if possible.

So, the primary thing that's changed is that all types now implement `Encode`, which allows us to ensure at compile-time that the type's ABI is valid (e.g. it's not `Vec<u8>`, `&str` or similar), and at runtime that the type matches what is expected by the method.
I really should write some better docs on this, but I recently explained it [here](https://github.com/ryanmcgrath/cacao/pull/30#issuecomment-1660162193), that may help - if not, feel free to ask!

Re compilation time, a clean build has not changed much, we've traded the `core-graphics-types` crate for `icrate`, and both of those compile faster than `syn` anyhow. But `icrate` is big, and not as optimized on that front as it should be, so there's definitely still room for improvement.

---

Note: This PR _is_ a breaking change, both because of the changed `Message` impls for everything (e.g. code that wanted to send messages manually to `metal` types now have to use `objc2`, or have to do a cast), but also because I changed a lot of APIs to no longer directly depend on Objective-C types. Concretely, I made the following public-facing changes:
- Removed `NSInteger` and `NSUInteger` type aliases, instead we just use `isize` and `usize`.
- Removed `NSRange`, methods now take `Range<usize>` instead.
- Use `f64` in `[set_]contents_scale` instead of `CGFloat`.
- Changed the signature of `drawable_size` from `fn(&self) -> CGSize` to `fn(&self) -> (f64, f64)`.
- Changed the signature of `set_drawable_size` from `fn(&self, size: CGSize)` to `fn(&self, width: f64, height: f64)`.

src/lib.rs
```diff
@@ -284,51 +244,84 @@ where
...
         }
     }
+}
 
+impl<T> Array<T>
+where
+    T: ForeignType + objc2::Encode + 'static,
+    T::Ref: objc2::Message + 'static,
+{
     pub fn from_owned_slice<'a>(s: &[T]) -> &'a ArrayRef<T> {
```
madsmtm, line 310: I'm unsure as to how this was supposed to work before? It seems like we're trying to pass `*const T`, which in this case could be e.g. `MetalLayer`, but `MetalLayer` itself contains a pointer, so ABI-wise we're actually passing a `*const *const CAMetalLayer`.

madsmtm: Nvmd, I fixed it now.

madsmtm:
Just tested, all the examples except the ones in https://github.com/gfx-rs/metal-rs/issues/282 now work (the errors in that issue are caught by debug assertions, which is why they don't).

madsmtm:
@grovesNL: I'm a bit unsure because I think you asked about a lot of things that are not strictly relevant to this PR anymore (since I've kept the autogenerated bindings out for now), but think I've resolved/answered your questions, are there any outstanding concerns, or something else?

CC @kvark, @cwfitzgerald

zackangelo:
Is there anything blocking the merge of this PR? I'm happy to help if needed.

cwfitzgerald:
Closing this PR, as we want people to use objc2-metal going forward as opposed to this crate. See #339

PR #242, opened 2022-09-05
Add `CGColorSpace` type

jinleili:
Prepare to add color space to wgpu

### DIFF
--- a/src/color_space.rs
+++ b/src/color_space.rs
@@ -5,21 +5,55 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use objc::runtime::{BOOL, YES};
+
 #[repr(C)]
 pub struct __CFString(std::ffi::c_void);
 pub type CFStringRef = *const __CFString;
 
+#[derive(Debug, Clone, Copy)]
 pub enum ColorSpaceName {
-    SRGB,
+    Srgb,
+    SrgbExtended,
+    SrgbExtendedLinear,
+    SrgbLinear,
+
     DisplayP3,
-    ExtendedLinearSRGB,
-    ExtendedDisplayP3,
-    ExtendedLinearDisplayP3,
+    DisplayP3Extended,
+    DisplayP3ExtendedLinear,
+    DisplayP3Linear,
+
+    BT2020,
+    BT2020SrgbGamma,
+    BT2020Extended,
+    BT2020ExtendedLinear,
     BT2020Linear,
-    BT2020PQ,
-    BT2020HLG,
     BT2100PQ,
     BT2100HLG,
+}
+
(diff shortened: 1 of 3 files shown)


src/color_space.rs
```diff
@@ -0,0 +1,83 @@
...
+    // CG_AVAILABLE_STARTING(10.14.3, 12.3)
+    static kCGColorSpaceExtendedLinearDisplayP3: CFStringRef;
+
+    // HDR
+    // CG_AVAILABLE_STARTING(12.0, 15.0)
+    static kCGColorSpaceLinearITUR_2020: CFStringRef;
+    // CG_AVAILABLE_BUT_DEPRECATED(10.15.4, 11.0, 13.4, 14.0);
+    static kCGColorSpaceITUR_2020_PQ: CFStringRef;
+    // CG_AVAILABLE_BUT_DEPRECATED(10.15.6, 11.0, 12.6, 14.0)
+    static kCGColorSpaceITUR_2020_HLG: CFStringRef;
```
cwfitzgerald, line 75: Let's get rid of these, There's no difference between these and the 2100 ones, the 2020 ones were just mildly incorrectly named.

src/color_space.rs
```diff
@@ -0,0 +1,83 @@
...
+    // CG_AVAILABLE_STARTING(10.12, 10.0)
+    static kCGColorSpaceExtendedLinearSRGB: CFStringRef;
+    // CG_AVAILABLE_STARTING(11.0, 14.0)
+    static kCGColorSpaceExtendedDisplayP3: CFStringRef;
+    // CG_AVAILABLE_STARTING(10.14.3, 12.3)
+    static kCGColorSpaceExtendedLinearDisplayP3: CFStringRef;
+
+    // HDR
+    // CG_AVAILABLE_STARTING(12.0, 15.0)
+    static kCGColorSpaceLinearITUR_2020: CFStringRef;
```
cwfitzgerald, line 71: We seem to be missing https://developer.apple.com/documentation/coregraphics/cgcolorspace/3892090-itur_2020_srgbgamma, which does come up

cwfitzgerald (changes requested):
Looking good, some comments.

madsmtm (commented):
Just noting: Somewhat duplicate of what's in https://github.com/servo/core-foundation-rs/blob/de8752262a2f4a253c8c0f214294094c28e9c0ba/core-graphics/src/color_space.rs

jinleili:
Yes, but metal-rs currently does not directly depend on `core-graphics` crate, just `core-graphics-types`

AlexanderEkdahl:
@jinleili I added these color spaces in https://github.com/servo/core-foundation-rs/pull/533 which would allow your implementation to use extended range rendering if you depend on `core-graphics`.

PR #243, opened 2022-09-20
added convenience, build buffer from slice, leveraging rust generics …

dobkeratops:
.. I realise the intent is probably to give an API that is a 1:1 translation of the objC/swift originals, but I'm guessing almost everyone is going to write this buffer constructor helper, and this kind of streamlining makes the example easier to comprehend.

any thoughts?

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1965,6 +1965,18 @@ impl DeviceRef {
                                        options:options]
         }
     }
+    
+    pub fn new_buffer_from_slice<T:Copy>(
+        &self,
+        data:&[T],
+        options:MTLResourceOptions 
+    )-> Buffer {
+        self.new_buffer_with_data(
+            &data[0] as *const _ as *const std::ffi::c_void,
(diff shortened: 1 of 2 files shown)


src/device.rs
```diff
@@ -1965,6 +1965,18 @@ impl DeviceRef {
...
         }
     }
+    
+    pub fn new_buffer_from_slice<T:Copy>(
+        &self,
+        data:&[T],
+        options:MTLResourceOptions 
+    )-> Buffer {
+        self.new_buffer_with_data(
+            &data[0] as *const _ as *const std::ffi::c_void,
```
madsmtm, line 1975: Use `slice::as_ptr`, otherwise you won't get the correct provenance!

```suggestion
            data.as_ptr().cast::<std::ffi::c_void>(),
```

madsmtm (changes requested):

cwfitzgerald (approved):
Needs formatting, but good after

PR #250, opened 2022-12-10
Host documentation

IsaccBarker:
See issue #5, specifically [here](https://github.com/gfx-rs/metal-rs/issues/5#issuecomment-1345299215).

GitHub Pages still has to be set up via repository settings, but this should work. It worked on [my fork](https://isaccbarker.github.io/metal-rs/metal/).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,42 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Document
+      run: cargo doc --verbose
+    - name: Packaging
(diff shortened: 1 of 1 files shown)


kvark:
@IsaccBarker thank you! What changes need to be done, exactly? Pages enabled in what way?

IsaccBarker:
If I recall correctly from when I set it up on my own repository, you should be able to go to the pages tab in the repository settings page, and select the source to be from Github Actions.

Skærmbillede 2022-12-25 kl  11 53 27 PM

You can see above that the docs building succeeded, but the deploy failed. I *believe* (I don't have access to the logs), this is because the source isn't set to Github Pages. If you change that, I think it should work.

Thanks!

kvark:
Thanks for the info! I switched the github pages source to github actions, but the CI action still seems to be failing. Anything I'm missing?

.github/workflows/rust.yml
```diff
@@ -0,0 +1,42 @@
+name: Rust
```
MarijnS95, line 1:
```suggestion
name: Publish Rust documentation to GitHub pages
```

And perhaps rename the file as well?

MarijnS95 (changes requested):
Can this be retried? It's quite annoying to have to find & jump to a custom fork to read the documentation.

Also don't forget to update `Cargo.toml` with a `documentation=` link to GitHub pages, and improve the mentions in the readme.

MarijnS95:
How about updating the PR description with `Fixes #5 fixes #276`?

PR #252, opened 2022-12-23
Adding support for the raytracing.

Hossein-Noroozpour:


src/accelerator_structure.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub fn descriptor() -> Self {
+        unsafe {
+            let class = class!(MTLPrimitiveAccelerationStructureDescriptor);
+            msg_send![class, descriptor]
+        }
+    }
+}
+
+impl PrimitiveAccelerationStructureDescriptorRef {
+    pub fn set_geometry_descriptors(&mut self, descriptors: &ArrayRef<AccelerationStructureGeometryDescriptor>) {
```
kvark, line 57: why do we have `&mut self`? Since `self` is a ref type, requiring unique borrowing doesn't protect against anything. In general, unique borrowing in metal-rs isn't functional, it can't be enforced, unfortunately.

src/lib.rs
```diff
@@ -453,33 +480,35 @@ impl MetalLayerRef {
...
-mod mps;
-mod pipeline;
-mod renderpass;
-mod resource;
-mod sampler;
-mod sync;
-mod texture;
-mod types;
-mod vertexdescriptor;
+pub mod mps;
```
kvark, line 499: why are we making everything pub?
We should either have everything private and re-exported, or everything public and visible in the module.
Unless there is a good reason to change this, we should default to just keep the existing approach, where the modules are private.

kvark (commented):
This is amazing work ❤️ , and a great present for New Year / Christmas to the community!
I was just thinking about how it would be required for metal-rs to get the RT support before we can get anywhere with a portable RT API/implementation, and your PR couldn't be timed more perfectly.
My only concern is how this giant change got unnecessarily more giant by including the foreign types update (long-needed!). Ideally, the foreign-types would be updated in a separate commit (or a PR). I don't consider this a blocker, though.
Let's just resolve the few notes raised and proceed (squashed).

examples/raytracing/geometry.rs
```diff
@@ -0,0 +1,351 @@
...
+#[derive(Default)]
+#[repr(C)]
+pub struct Triangle {
+    pub normals: [Vec4; 3],
+    pub colours: [Vec4; 3],
+}
+
+pub fn get_managed_buffer_storage_mode() -> MTLResourceOptions {
+    return MTLResourceOptions::StorageModeManaged;
+    // return MTLResourceStorageModeShared;
```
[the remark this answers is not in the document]
Hossein-Noroozpour: :sweat: Ah, sorry it's left, actually, I wanted to support iOS too, but thought it may make the example code harder to read because of that I didn't add the iOS codes and this code left.

src/accelerator_structure.rs
Hossein-Noroozpour: Actually, in the first place, it was my question, why don't we have the unique borrowing for the codes that actually are mutating the inner state of a Metal object, here I thought it might bring more safety to the code.

But however, as you say it "isn't functional" and I'll remove it.

src/lib.rs
Hossein-Noroozpour: The reason is, we have name-clashes in other modules, for example in `mps` we have `AccelerationStructure` and in `accelerator_structure` too.
The interesting thing is when I was developing it I was using nightly and it was able to distinguish them and use the right type, but stable wasn't able to find the right type because of that name-clash.

kvark: thanks for clarification! Would it be fine to just make `mps` a public module, while reexporting everything else at root? After all, mps is really weird because it's a separate library at the system level. We may even gate it by a feature.

src/accelerator_structure.rs
kvark: Maybe this needs clarification, so here it is. Metal relies on objc and foreign-types. These allow one to create an object and then create references to this object. A reference borrows the object. So if you have a `&mut self` method, you aren't going to be able to call if it there are any `MyObjectRef` around, including calling it on a reference.
The other problem is that nothing in objc guarantees that an object is unique. I.e. one can get a reference from an object, and then do `to_owned`, and get another owned object that represents the same metal object.

TL;DR: objc/foreign-types provide no uniqueness guarantees, therefore `&mut self` methods aren't sound.

kvark (approved):

kvark:
Let's proceed when you are ready.

src/lib.rs
Hossein-Noroozpour: Yeah, it's a great idea! So I'm going to remove the `pub use mps::*;` and only keep the `mps` pub.

Hossein-Noroozpour: Done

Hossein-Noroozpour:
It's ready!

PR #253, opened 2023-01-16
Add `is_null()` helper function to Texture

jinleili:
Avoid adding direct dependencies of the `foreign_types` crate in wgpu to determine if the Texture is NULL.

wgpu issue: https://github.com/gfx-rs/wgpu/issues/3153

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -352,4 +352,14 @@ impl TextureRef {
     pub fn gpu_resource_id(&self) -> MTLResourceID {
         unsafe { msg_send![self, gpuResourceID] }
     }
+
+    pub fn is_null(&self) -> bool {
+        use foreign_types::ForeignTypeRef;
+
+        if self.as_ptr().is_null() {
+            true
+        } else {
+            false
+        }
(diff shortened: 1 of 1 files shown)


src/texture.rs
```diff
@@ -352,4 +352,14 @@ impl TextureRef {
     pub fn gpu_resource_id(&self) -> MTLResourceID {
         unsafe { msg_send![self, gpuResourceID] }
     }
+
+    pub fn is_null(&self) -> bool {
+        use foreign_types::ForeignTypeRef;
+
+        if self.as_ptr().is_null() {
```
kvark, line 359: do we really need `if xxx { true} else { false }` ? why not just `xxx`?

jinleili: Sorry, since `foreign_types` crate has been introduced as a dependency of wgpu, there is no longer any reason to add this function.

jinleili:
Sorry, since `foreign_types` crate has been introduced as a dependency of wgpu, there is no longer any reason to add this function.

PR #256, opened 2023-02-24
Clean up RT API

kvark:
Follow-up to #252 
cc @Hossein-Noroozpour


kvark:
Contains a bunch of essential fixes, such as function signature ones, and more complete set of functions.
So I'll go ahead and merge.

PR #263, opened 2023-03-29
feature: Metal timestamps

FL33TW00D:
This is required for wgpu to ship timestamp queries on Metal.

Related Issue: https://github.com/gfx-rs/wgpu/issues/2739

- [X] add functionality to render passes.
- [X] resolve to actual time instead of internal time.
- [X] correctly fetch and extract timestamp counterset from counterset array.


examples/counters/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    let range = crate::NSRange::new(0_u64, 2_u64);
+    blit_encoder.resolve_counters(&counter_sample_buffer, range, &destination_buffer, 0_u64);
+    blit_encoder.end_encoding();
+
+    command_buffer.commit();
+    command_buffer.wait_until_completed();
+
+    let timestamps =
+        unsafe { std::slice::from_raw_parts(destination_buffer.contents() as *const u64, 2) };
+    println!("Start timestamp: {}", timestamps[0]);
```
kvark, line 91: are these really timestamps? AFAIK they are just abstract sample counts

src/device.rs
```diff
@@ -2149,4 +2163,22 @@ impl DeviceRef {
     ) -> accelerator_structure::AccelerationStructure {
         unsafe { msg_send![self, newAccelerationStructureWithSize: size] }
     }
+
+    pub fn sample_timestamps(&self, cpu_timestamp: &mut u64, gpu_timestamp: &mut u64) {
+        unsafe { msg_send![self, sampleTimestamps: cpu_timestamp gpuTimestamp: gpu_timestamp] }
+    }
+
+    pub fn counter_sets(&self) -> Vec<CounterSet> {
```
kvark, line 2171: We already have an abstraction over NSArray. Should we just return it instead of manually collecting into `Vec`?

kvark (commented):
Very nice!

examples/counters/main.rs
```diff
@@ -0,0 +1,151 @@
+use metal::*;
+fn main() {
+    let device = Device::system_default().expect("No device found");
```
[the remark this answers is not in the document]
kvark: which means there is likely a legitimate issue with the refcounts in timestamp code
merging would be nice! I don't like having hundreds of small examples, it's too much to maintain

FL33TW00D:
@kvark Ready to go now - I fixed the segfault and merged the `compute` and `counters` example.

kvark (approved):

PR #264, opened 2023-03-30
Add support for object and mesh shader

zmarlon:
I added support for Object and Mesh Shaders including a short example.


src/encoder.rs
```diff
@@ -440,6 +440,278 @@ impl RenderCommandEncoderRef {
...
+            ]
+        }
+    }
+
+    /// Only available in (macos(13.0), ios(16.0))
+    pub fn set_object_sampler_state_with_lod(
+        &self,
+        index: NSUInteger,
+        sampler: Option<&SamplerStateRef>,
+        lod_clamp: Range<f32>,
```
kvark, line 502: I think the use of `Range` here is slight off from the semantics. Generally `Range` is excluding the top bound (while including the lower bound), while the LOD clamping semantics is inclusive from both sides.
Perhaps we should just have this function accept min and max as separate values?

src/encoder.rs
```diff
@@ -440,6 +440,278 @@ impl RenderCommandEncoderRef {
...
+            ]
+        }
+    }
+
+    /// Only available in (macos(13.0), ios(16.0))
+    pub fn set_mesh_sampler_state_with_lod(
+        &self,
+        index: NSUInteger,
+        sampler: Option<&SamplerStateRef>,
+        lod_clamp: Range<f32>,
```
kvark, line 643: same here

kvark (commented):
Very nice! Just one small thing I see

src/encoder.rs
zmarlon: Since `set_fragment_sampler_state_with_lod` and others also uses a `Range`, I would also stick with this for object and mesh shaders for consistency. Otherwise we would have to change it all over the place and make a breaking change in the API. What do you prefer?

kvark: fair point, we were wrong to begin with, but your PR is not the place to fix it

kvark (approved):

PR #265, opened 2023-04-01
chore: Fix timestamp bug and add required method

FL33TW00D:
Introduced a bug in the last PR.

Looks like everything should be good to go here, the retain is no longer needed thanks to @Wumpf.

### DIFF
--- a/src/computepass.rs
+++ b/src/computepass.rs
@@ -72,7 +72,10 @@ impl ComputePassSampleBufferAttachmentDescriptorRef {
     }
 
     pub fn set_sample_buffer(&self, sample_buffer: &CounterSampleBufferRef) {
-        unsafe { msg_send![self, setSampleBuffer: sample_buffer] }
+        unsafe {
+            let () = msg_send![sample_buffer, retain];
+            msg_send![self, setSampleBuffer: sample_buffer]
+        }
     }
 
     pub fn start_of_encoder_sample_index(&self) -> NSUInteger {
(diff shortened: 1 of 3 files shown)


src/computepass.rs
```diff
@@ -72,7 +72,10 @@ impl ComputePassSampleBufferAttachmentDescriptorRef {
     }
 
     pub fn set_sample_buffer(&self, sample_buffer: &CounterSampleBufferRef) {
-        unsafe { msg_send![self, setSampleBuffer: sample_buffer] }
+        unsafe {
+            let () = msg_send![sample_buffer, retain];
```
FL33TW00D, line 76: I needed this `retain` to stop SEGFAULTs during wgpu implementation.
Unsure of it's validity.

https://developer.apple.com/documentation/metal/mtlcomputepasssamplebufferattachmentdescriptor/3564439-samplebuffer?language=objc

madsmtm: `retain` returns a pointer, so you should do `let _: *const Object = msg_send![sample_buffer, retain]`.

Not entirely sure why you have memory management problems, but an extra `retain` is always sound (it'll just leak if incorrect).

kvark: @FL33TW00D this looks like it could be a problem on `wgpu` side. Metal shouldn't be explicitly adding this `retain`, and instead `wgpu` should keep the object alive, just like it does for any other resources referenced by command buffers.

kvark (changes requested):

kvark (approved):

PR #266, opened 2023-04-11
Don't release the error obj manually

teoxoy:
We should do this for the same reason it was done in: https://github.com/gfx-rs/metal-rs/pull/196 (calling functions which use this macro in an `autorelasepool` will cause a SIGSEGV since the obj has already been released manually).

Fixes the segfault encountered in: https://github.com/gfx-rs/wgpu/issues/3550.


kvark (approved):
nice find!

PR #267, opened 2023-04-15
Remove unreachable code

AlexanderEkdahl:
Not sure what exactly changed (Rust version?, dependency version?) since I haven't seen this when working on this codebase earlier.

### DIFF
--- a/src/argument.rs
+++ b/src/argument.rs
@@ -270,7 +270,6 @@ impl ArgumentRef {
             match msg_send![self, isActive] {
                 YES => true,
                 NO => false,
-                _ => unreachable!(),
             }
         }
     }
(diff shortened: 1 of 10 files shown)


madsmtm:
This only works on `aarch64`, on `x86_64` the `YES` and `NO` constants are different.

AlexanderEkdahl:
> This only works on `aarch64`, on `x86_64` the `YES` and `NO` constants are different.

Oh I see. Thank you for the insight. 

I'm trying to think of other workarounds to remove the linter noise.

AlexanderEkdahl:
One possible workaround is of course something like the following:

``` rust
match msg_send![self, isFramebufferOnly] {
    YES => true,
    NO => false,
    #[cfg(not(target_arch = "aarch64"))]
    _ => unreachable!(),
}
```

fornwall:
> One possible workaround is of course something like the following:
> 
> ```rust
> match msg_send![self, isFramebufferOnly] {
>     YES => true,
>     NO => false,
>     #[cfg(not(target_arch = "aarch64"))]
>     _ => unreachable!(),
> }
> ```

Proposed that in #273.

AlexanderEkdahl:
Thank you @fornwall

PR #269, opened 2023-05-28
Bump `bitflags` to v2

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented by the macro is now different.

I was a bit liberal with deriving `std` traits on public types, which mimics the behavior of `bitflags` v1, let me know if we don't need all these traits. Potentially we could at least drop `PartialOrd` and `Ord`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ mps = []
 
 [dependencies]
 core-graphics-types = "0.1"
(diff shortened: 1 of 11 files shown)


daxpedda:
CI is broken with the same error as on master.

PR #273, opened 2023-07-05
Fix unreachable pattern warnings on aarch64

fornwall:
This fixes `warning: unreachable pattern` on `aarch64`, as seen when running e.g. `cargo clippy` or `cargo run --example compute`.

As suggested by [this comment](https://github.com/gfx-rs/metal-rs/pull/267#issuecomment-1509633887) in #267 (which this PR replaces).

### DIFF
--- a/src/argument.rs
+++ b/src/argument.rs
@@ -270,6 +270,7 @@ impl ArgumentRef {
             match msg_send![self, isActive] {
                 YES => true,
                 NO => false,
+                #[cfg(not(target_arch = "aarch64"))]
                 _ => unreachable!(),
             }
         }
--- a/src/argument.rs
+++ b/src/argument.rs
@@ -266,14 +266,7 @@ impl ArgumentRef {
     }
 
     pub fn is_active(&self) -> bool {
-        unsafe {
-            match msg_send![self, isActive] {
-                YES => true,
-                NO => false,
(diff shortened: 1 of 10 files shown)


fornwall:
I think the build failures should be fixed by #275.

src/argument.rs
```diff
@@ -270,6 +270,7 @@ impl ArgumentRef {
             match msg_send![self, isActive] {
                 YES => true,
                 NO => false,
+                #[cfg(not(target_arch = "aarch64"))]
```
kvark, line 273: can we just allow the "_" specifically across the project on `aarch64`  instead of patching every single match?

fornwall: @kvark I tried a variant with macros to remove repetition - what do you think? https://github.com/gfx-rs/metal-rs/pull/273/commits/9618db112452a5a171092b59b1002a3d06af63b4

kvark (approved):
looks better, thank you!

src/argument.rs
kvark: This is not exactly what I wanted, but it's definitely better.
What I was thinking is - take that warning and put something like this at the top of `lib.rs`:
```rust
#![cfg_attr(target_arch = "aarch64", allow(this_nasty_warning))]
```

kvark:
Please proceed when you feel ready

PR #274, opened 2023-07-05
Improve the compute example to work on large array sizes

### LINKED ISSUE #244: Incorrect buffer value in compute example
Hi,
I was trying out some of the examples and it looks like the `compute` example only works for small vector sizes. To reproduce the issue simply replace the `let data = [1u32, 2, 3 , etc.. ];` to `let data = vec![1u32; 100000];` and assert that the sum should be 100000. In my case the value in the sum buffer is always the unmodified value 0. 
Can anyone explain to me what is happening here?
Best,
Philippe

fornwall:
Improve the compute example to work on large array sizes by starting from the thread group size and calculate the thread group count, instead of the other way around (which can cause too large thread groups). This fixes #244.

Also make it possible to specify array size as argument and fix clippy warnings.

### DIFF
--- a/examples/compute/main.rs
+++ b/examples/compute/main.rs
@@ -3,9 +3,13 @@ use objc::rc::autoreleasepool;
 use std::path::PathBuf;
 
 const NUM_SAMPLES: u64 = 2;
-const NUM_ELEMENTS: u64 = 64 * 64;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


fornwall:
I think the build failures should be fixed by #275.

kvark (approved):

PR #276, opened 2023-07-14
build(docs.rs): cover more Mac targets

### LINKED ISSUE #5: Host documentation

ErichDonGubler:
Expands the set of Mac targets to cover more Mac toolchains. 😀

<details><summary>Original OP text from draft</summary>

We've specified a `default-target` for `docs.rs` metadata in `Cargo.toml`, but the way to exclude other platforms is via the `targets` field, according to [these docs][targets-docs] on it:

> ```
> # If `default-target` is unset, the first element of `targets` is treated as the default target.
> # Otherwise, these `targets` are built in addition to the default target.
> # If both `default-target` and `targets` are unset,
> #   all tier-one targets will be built and `x86_64-unknown-linux-gnu` will be used as the default target.
> ```

[targets-docs]: https://web.archive.org/web/20230528205129/https://docs.rs/about/metadata

---

~~Should fix #5.~~ EDIT: [Maybe not?](https://github.com/gfx-rs/metal-rs/pull/276#issuecomment-1636313550)

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,13 @@ edition = "2021"
 exclude = ["guide/**/*", "examples/texture/**/*", "tests/**/*", "Cargo.lock", "target/**/*"]
 
 [package.metadata.docs.rs]
-default-target = "x86_64-apple-darwin"
+targets = [
+  "aarch64-apple-darwin", # presented first in Docs.rs, keep this here
+  "aarch64-apple-ios",
+  "aarch64-apple-ios-sim",
+  "x86_64-apple-darwin",
+  "x86_64-apple-ios",
+]
 
 [features]
 default = ["link"]


ErichDonGubler:
Ah, hmm, after reading through #5 a bit more, I'm not sure that this will fix things, since apparently Docs.rs cross-compiles from Linux, which might not have everything necessary to compile `metal`. It _should_ be more correct (because MacOS is the only platform for which this crate makes sense), but IDK if it will actually solve #5 now. 😕

MarijnS95:
Barring the cross-compilation issue, should iOS also be targeted for docs?

I'm unsure if there's anything `cfg`'d for that target, but might be useful to include it nonetheless.

Cargo.toml
```diff
@@ -13,7 +13,7 @@ edition = "2021"
 exclude = ["guide/**/*", "examples/texture/**/*", "tests/**/*", "Cargo.lock", "target/**/*"]
 
 [package.metadata.docs.rs]
-default-target = "x86_64-apple-darwin"
+targets = ["x86_64-apple-darwin"]
```
xorza, line 16: also `aarch64-apple-darwin`
and probably

```
aarch64-apple-ios
aarch64-apple-ios-sim
x86_64-apple-ios
```

MarijnS95:
@ErichDonGubler can you pursue this again? #5 was fixed and closed via #303, and running `cargo doc --open --no-deps --target x86_64-apple-darwin` locally shows that the docs now build successfully! Don't forget about my and @xorza's query to extend the list of targets first.

MarijnS95:
After all #303 was contained in the recent 0.28.0 release, and the default `x86_64-apple-darwin` target had a successful build **already** as seen at https://docs.rs/metal/0.28.0/metal/index.html!

ErichDonGubler:
@xorza: I've adjusted PR title, summary, and the sole commit to note that this only covers more targets, since that seems like the only useful scope left in this contribution. 🙂

ErichDonGubler:
@xorza: I've rebased, taken your review suggestion, adjusted PR title, summary, and the sole commit to note that this only covers more targets, since that seems like the only useful scope left in this contribution. 🙂

ErichDonGubler:
Nightly CI appears to be complaining about an "unused" feature `dispatch_queue`, which got renamed waaay back in https://github.com/gfx-rs/metal-rs/commit/fb17c9eba6fa619500c559f5a4526c4d96746ef7. Eek! That means the `dispatch` module has been disabled for…a long time, while a different route to `dispatch` functionality has been taken instead (maybe a private `use` from `lib.rs`?) and kept from breaking compilation.

MarijnS95 (approved):
Looks good to have all Tier 1/2 Apple platforms listed, and instead of setting a "useless" `default-target`, no more rely on the default `targets = []` array on which docs.rs will fail anyway: https://docs.rs/crate/metal/0.28.0/builds/1204445/x86_64-unknown-linux-gnu.txt

grovesNL (approved):
Thank you!

PR #281, opened 2023-08-25
Fix compiling for 32bit architectures

madsmtm:
A small slip-up that prevents compiling on 32bit architectures (which, admittedly, are rare nowadays).

I haven't tested the examples, there are probably similar mistakes as this one in there.

### DIFF
--- a/src/texture.rs
+++ b/src/texture.rs
@@ -117,7 +117,7 @@ impl TextureDescriptorRef {
             height,
             depth,
         } = size;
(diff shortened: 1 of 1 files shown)


madsmtm:
Converted to draft as this patch is only enough to fix `v0.25.0`; need to do more to fix on the current `master`.

cwfitzgerald:
Closing, as this repo is in maintenance mode.

PR #285, opened 2023-10-18
Matrix multiplication support

ivarflakstad:
Did this over the weekend. Far from finished (ref draft), but I wanted to get some early feedback 😊

I made some extra effort using traits to avoid improper usage for example
```rust
impl Valid for MatMulSpecification<Float32, Float16, Float32> {}
```

Let me know what you think!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ block = "0.1.6"
 foreign-types = "0.5"
 dispatch = { version = "0.2", optional = true }
 paste = "1"
+half = "2.3.1"
 
 [dependencies.objc]
 version = "0.2.4"
--- a/examples/mps/main.rs
+++ b/examples/mps/main.rs
@@ -1,3 +1,4 @@
+use metal::mps::*;
 use metal::*;
 use std::ffi::c_void;
 use std::mem;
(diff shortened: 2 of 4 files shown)


Narsil:
@cwfitzgerald (Sorry for the ping if you're not the correct person).

We're iterating quite a bit on this here -> https://github.com/ivarflakstad/metal-rs
We would love some feedback if upstreaming here makes sense or not.

ivarflakstad:
Fair enough 😊
Maybe you know who does?

EdorianDark:
@kvark Do you know who might be involved with metal-rs?

Narsil:
@Wumpf  Maybe ?

After having matured a bit this and implemented the `Matrix` in https://github.com/huggingface/candle/ I feel like this PR should be trimmed down to only get the `MPSMatrix` support and remove the rest of the sugar (being more aligned with the rest of the crate).

Ofc having maintainer advice would be helpful.
Also in order to be able to publish on our side, we're likely going to fork `metal-rs` for the time being to have access to those objects.

We might return to `metal-rs` once we get rid of `MPSMatrix` dependency (if we can using metal-flash-attention, which seems faster, but currently isn't  as widely portable as MPSMatrix since it doesn't work on first gen m1).
Either that or when this PR gets included.

PR #296, opened 2023-12-01
add `colorspace` to `MetalLayer`

ComfyFluffy:
Add methods to get and set `colorspace`. `core-graphics = "0.23.1"` is introduced.

https://developer.apple.com/documentation/quartzcore/cametallayer/1478170-colorspace?language=objc


kvark:
Ideally we avoid core-graphics dependency here...
https://github.com/servo/core-foundation-rs/issues/479

PR #298, opened 2023-12-13
Add `add_presented_handler` to `Drawable`

FlannyH:
Add the `add_presented_handler` function to `Drawable`
https://developer.apple.com/documentation/metal/mtldrawable/2806858-addpresentedhandler

PR #300, opened 2023-12-31
add `presented_time` to `Drawable` 

marcpabst:
https://developer.apple.com/documentation/metal/mtldrawable/2806855-presentedtime

In preparation for https://github.com/gfx-rs/wgpu/issues/2869.

### DIFF
--- a/src/accelerator_structure.rs
+++ b/src/accelerator_structure.rs
@@ -81,7 +81,9 @@ impl PrimitiveAccelerationStructureDescriptor {
     pub fn descriptor() -> Self {
         unsafe {
             let class = class!(MTLPrimitiveAccelerationStructureDescriptor);
-            msg_send![class, descriptor]
+            let ptr: *mut Object = msg_send![class, descriptor];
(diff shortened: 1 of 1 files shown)


MarijnS95:
Good to mention that this PR depends on #298, which should be (squash-)merged first.

Jasper-Bekkers:
And probably a nice idea to attribute the PR to the original author (@FlannyH)

marcpabst:
Yep - sorry! Never submitted a PR that supersedes another one and was under the impression that GitHub keeps track of that automatically.

MarijnS95:
@marcpabst no worries! GitHub doesn't have any support for this at all, so a little hint in the PR description is the least one can do (there's a finicky trick with target branches but it doesn't work across forks and is _super_ fragile).

Note also that you don't need to create a PR on your local fork to incorporate another change as a base - pro-tip for next time :+1:

grovesNL:
Just rebasing this, will merge in a minute

PR #307, opened 2024-04-22
ci: Update `actions/checkout` and `actions/cache`.

waywardmonkeys:
This updates these to the current versions which also use more recent versions of Node, silencing some deprecation warnings from within GitHub Actions.


grovesNL (approved):
Thank you!

Issue #308, opened 2024-04-22
new_library_with_data is broken


K0bin:
There appears to be a memory corruption bug with `new_library_with_data`.

Calling it results in all kinds of weird errors. Here are a few I've run into today while trying to figure out my problem:
- "readFunctionPublicOrPrivateMetadata:4956: failed assertion `Invalid library file/data (unexpected end of file)'"
- "Compiler encountered an internal error"
- "Function main0 does not exist" when trying to get a function from the library. And main0 does get listed when calling `function_names`, so the function definitely did exist.
- "assertion failed: !ptr.is_null()" (in metal-0.27.0/src/pipeline/render.rs:675:1)

Writing the exact same data to disk and using `new_library_with_file` instead works perfectly fine.

```rust
use metal;

const BROKEN: bool = true;

fn main() {
    println!("Hello, world!");

    let device = metal::Device::system_default().unwrap();
    let data = std::fs::read("test.metallib").unwrap();
    let data_box = data.into_boxed_slice();
    let lib = if BROKEN {
        device.new_library_with_data(&data_box).unwrap()
    } else {
        device.new_library_with_file("test.metallib").unwrap()
    };
    std::mem::drop(data_box);
    let function = lib.get_function("main0", None).unwrap();
    let _pipeline = device.new_compute_pipeline_state_with_function(&function).unwrap();
    println!("Done");
}
```
[Minimal reproducer that produces the failed assert inside the Metal driver with _with_data but not when using _with_file](https://github.com/gfx-rs/metal-rs/files/15067602/Reproducer.zip)

madsmtm:
Thanks for the reproducer, have posted https://github.com/gfx-rs/metal-rs/pull/309 to fix this.

PR #309, opened 2024-04-22
Fix unsoundness in `new_library_with_data`

madsmtm:
The `library_data` slice passed to the function that the dispatch data object is referencing does not necessarily outlive the library in which it will be contained.

(Note: It _looks_ like we're upholding memory management rules here, as the object returned from `dispatch_data_create` is released with `dispatch_release` before the end of the function, but remember that the dispatch data is a reference-counted object; `MTLLibrary` will retain the dispatch data beyond the lifetime of the function).

As specified in https://developer.apple.com/documentation/dispatch/1452970-dispatch_data_create, if we use DISPATCH_DATA_DESTRUCTOR_DEFAULT as the destructor block instead, `dispatch` will copy the buffer for us automatically.

Fixes https://github.com/gfx-rs/metal-rs/issues/308.


K0bin:
Thanks for the quick fix!

MarijnS95:
@madsmtm how about adding a comment to the code with your findings? Specifically, explain above `DISPATCH_DATA_DESTRUCTOR_DEFAULT,` that this will create a copy of the passed data so that the dispatch data becomes independent of `library_data: &[u8]`? Could that be part of `// Safety` docs on `unsafe` for the `dispatch_data_create()` function?

MarijnS95 (approved):

grovesNL (approved):
Thank you!

PR #312, opened 2024-04-24
Don't re-export `bitflags` and `log`.

waywardmonkeys:
Also, switch away from `#[macro_use]` and `extern crate` for these crates.


grovesNL (approved):
Thanks!

PR #314, opened 2024-05-01
Implement indirect instance acceleration structure

FlannyH:
Implements the `MTLIndirectAccelerationStructureInstanceDescriptor` struct, the `MTLIndirectInstanceAccelerationStructureDescriptor` struct with all its setter functions. Also implements fetching an acceleration structure's GPU resource id, since we need that to fill in the `MTLIndirectAccelerationStructureInstanceDescriptor`


kvark (approved):
nice!

PR #315, opened 2024-05-01
Implement Visible Function Table

FlannyH:
Implements `MTLVisibleFunctionTableDescriptor`, creating a visible function table on a pipeline, as well as the different `setVisibleFunctionTable` functions in the command encoders.

### DIFF
--- a/src/accelerator_structure.rs
+++ b/src/accelerator_structure.rs
@@ -572,3 +572,51 @@ impl IntersectionFunctionTableRef {
         unsafe { msg_send![self, gpuResourceID] }
     }
 }
+
+// See https://developer.apple.com/documentation/metal/mtlvisiblefunctiontabledescriptor/
+pub enum MTLVisibleFunctionTableDescriptor {}
+
+foreign_obj_type! {
+    type CType = MTLVisibleFunctionTableDescriptor;
+    pub struct VisibleFunctionTableDescriptor;
+    type ParentType = NsObject;
+}
+
+impl VisibleFunctionTableDescriptor {
+    pub fn new() -> Self {
+        unsafe {
+            let class = class!(MTLVisibleFunctionTableDescriptor);
+            msg_send![class, new]
+        }
+    }
+}
+
+impl VisibleFunctionTableDescriptorRef {
+    pub fn set_function_count(&self, count: NSUInteger) {
+        unsafe { msg_send![self, setFunctionCount: count] }
+    }
+}
+
+// See https://developer.apple.com/documentation/metal/mtlvisiblefunctiontable
+pub enum MTLVisibleFunctionTable {}
+
+foreign_obj_type! {
+    type CType = MTLVisibleFunctionTable;
+    pub struct VisibleFunctionTable;
+    type ParentType = Resource;
+}
+
+impl VisibleFunctionTableRef {
+    pub fn set_functions(&self, functions: &[&FunctionRef]) {
+        let ns_array = Array::<Function>::from_slice(functions);
+        unsafe { msg_send![self, setFunctions: ns_array] }
+    }
+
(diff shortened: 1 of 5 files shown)


src/pipeline/render.rs
```diff
@@ -688,6 +688,17 @@ impl RenderPipelineStateRef {
...
     }
+    /// Only available on (macos(11.0), ios(14.0))
+    pub fn new_visible_function_table_with_descriptor(
+        &self,
+        descriptor: &VisibleFunctionTableDescriptorRef,
+        stage: MTLRenderStages,
+    ) -> VisibleFunctionTable {
+        unsafe {
+            msg_send![self, newVisibleFunctionTableWithDescriptor: descriptor
+                                                                            stage:stage]
```
MarijnS95, line 699: Did `rustfmt` do this?

src/pipeline/render.rs
```diff
@@ -688,6 +688,17 @@ impl RenderPipelineStateRef {
             crate::nsstring_as_str(label)
         }
     }
+    /// Only available on (macos(11.0), ios(14.0))
```
MarijnS95, line 691:
```suggestion

    /// Only available on (macos(11.0), ios(14.0))
```

src/accelerator_structure.rs
```diff
@@ -346,3 +346,51 @@ impl IntersectionFunctionTableRef {
         unsafe { msg_send![self, setFunction: function atIndex: index] }
     }
 }
+
+// See https://developer.apple.com/documentation/metal/mtlvisiblefunctiontabledescriptor/
```
MarijnS95, line 350: Rustdoc and formatting:
```suggestion
/// See <https://developer.apple.com/documentation/metal/mtlvisiblefunctiontabledescriptor>
```

src/accelerator_structure.rs
```diff
@@ -346,3 +346,51 @@ impl IntersectionFunctionTableRef {
...
+    }
+}
+
+impl VisibleFunctionTableDescriptorRef {
+    pub fn set_function_count(&self, count: NSUInteger) {
+        unsafe { msg_send![self, setFunctionCount: count] }
+    }
+}
+
+// See https://developer.apple.com/documentation/metal/mtlvisiblefunctiontable
```
MarijnS95, line 374:
```suggestion
/// See <https://developer.apple.com/documentation/metal/mtlvisiblefunctiontable>
```

src/pipeline/render.rs
FlannyH: Fixed in latest commit

src/accelerator_structure.rs
```diff
@@ -345,4 +345,80 @@ impl IntersectionFunctionTableRef {
     pub fn set_function(&self, function: &FunctionHandleRef, index: NSUInteger) {
         unsafe { msg_send![self, setFunction: function atIndex: index] }
     }
+
+    pub fn set_visible_function_table(
+        &self,
+        visible_function_table: Option<&VisibleFunctionTableRef>,
+        buffer_index: NSUInteger,
```
kvark, line 352: there is a bit of a mismatch with the code we have right now, see https://github.com/gfx-rs/metal-rs/blob/3b57d7164f2702acd8a78aa3ed1131c1886a993a/src/encoder.rs#L299

In Rust (and just common sense) you'd expect this to work essentially as `container[buffer_index] = something`, so you specify the buffer index first. However, in Metal it's last for some reason. Previously, we'd acklowledge this as being one of the points where we derive from Metal, but it's not set in stone - we are open to reconsider this.

So, perhaps you'd be open to put the `buffer_index` first in those methods?

FlannyH: Yeah that seems better. I changed it just now in the latest commit so the index is first

PR #316, opened 2024-05-01
Implement creating acceleration structures on a `Heap`

FlannyH:
Implements allocating acceleration structures from a heap using the `newAccelerationStructureWithDescriptor` and `newAccelerationStructureWithSize` functions, as well as `heapAccelerationStructureSizeAndAlignWithSize` on Device. Additionally implements two missing functions `setHazardTrackingMode` and `setType`.

### DIFF
--- a/src/heap.rs
+++ b/src/heap.rs
@@ -148,6 +148,69 @@ impl HeapRef {
             }
         }
     }
+
+    /// Only available on macOS 13.0+ & iOS 16.0+
+    pub fn new_acceleration_structure_with_descriptor(
+        &self,
+        descriptor: &AccelerationStructureDescriptorRef,
+    ) -> Option<AccelerationStructure> {
+        unsafe {
+            let ptr: *mut MTLAccelerationStructure =
+                msg_send![self, newAccelerationStructureWithDescriptor: descriptor];
+            if !ptr.is_null() {
+                Some(AccelerationStructure::from_ptr(ptr))
+            } else {
+                None
+            }
+        }
+    }
+
+    /// Only available on macOS 13.0+ & iOS 16.0+
+    pub fn new_acceleration_structure_with_descriptor_offset(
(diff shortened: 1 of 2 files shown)


src/heap.rs
```diff
@@ -206,4 +273,8 @@ impl HeapDescriptorRef {
     pub fn heap_type(&self) -> MTLHeapType {
         unsafe { msg_send![self, type] }
     }
+
+    pub fn set_heap_type(&self, type_: MTLHeapType) {
```
MarijnS95, line 277: Note that `objc2-metal` calls these just `type()`.

Also, does the `/// Only available on macos(10.15), ios(13.0)` comment apply here too?

src/heap.rs
```diff
@@ -197,6 +260,10 @@ impl HeapDescriptorRef {
         unsafe { msg_send![self, hazardTrackingMode] }
     }
 
+    pub fn set_hazard_tracking_mode(&self, hazard_tracking_mode: MTLHazardTrackingMode) {
```
MarijnS95, line 263: `/// Only available on macos(10.15), ios(13.0)`?

src/heap.rs
FlannyH: I decided to be consistent with the already existing function `fn heap_type()`, so I wouldn't have to change already existing code. Do we want to change the original function too?

FlannyH: Also added the doc comment for all functions I added in this PR with the right versions in latest commit

src/heap.rs
FlannyH: Added it in latest commit

robclu:
Hi, is there any reason this isn't merged? I would also like to be able to set the hazard tracking mode and heap type (for placement heaps).

grovesNL (approved):
Thank you!

PR #321, opened 2024-05-23
Add missing enum variants for MTLDataType

FlannyH:
Adds all the missing enum variants for MTLDataType, including 64 bit integers, pipelines, indirect command buffer, ray tracing related types, and BFloat.


Jasper-Bekkers (approved):

Jasper-Bekkers:
Nightly breakage seems unrelated.

PR #326, opened 2024-06-11
Correct `refit_acceleration_structure` destination parameter

FlannyH:
According to [the Metal documentation for `refitAccelerationStructure`](https://developer.apple.com/documentation/metal/mtlaccelerationstructurecommandencoder/3553898-refitaccelerationstructure?language=objc), we can pass nil to `destinationAccelerationStructure`, which will then refit the acceleration structure in-place. This PR turns the parameter into an `Option<_>`, so the user can pass `None` to refit in place.


grovesNL (approved):
Thank you!

PR #328, opened 2024-07-13
sampler: Add missing `gpu_resource_id()` getter

MarijnS95:
According to the upstream docs and the Metal Shader Converter implementation the `SamplerState` type has a way to get access to its `MTLResourceID` on the GPU: https://developer.apple.com/documentation/metal/mtlsamplerstate/3974101-gpuresourceid

### DIFF
--- a/src/sampler.rs
+++ b/src/sampler.rs
@@ -5,7 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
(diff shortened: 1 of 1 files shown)


madsmtm (changes requested):
The label is explicitly marked as `readonly` in [the documentation](https://developer.apple.com/documentation/metal/mtlsamplerstate/1516329-label?language=objc), so I wouldn't rely on it being settable after the creation of the sampler state.

cwfitzgerald (approved):

PR #329, opened 2024-07-15
Add `gpu_start_time` and `gpu_end_time` to `CommandBuffer`

zackangelo:
Add bindings for [`GPUStartTime`](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1639924-gpustarttime?language=objc) and [`GPUEndTime`](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1639926-gpuendtime?language=objc). 

They allow you to determine how much time the GPU spent executing a command buffer.

### DIFF
--- a/src/commandbuffer.rs
+++ b/src/commandbuffer.rs
@@ -66,6 +66,14 @@ impl CommandBufferRef {
         }
     }
 
+    pub fn gpu_start_time(&self) -> f64 {
+        unsafe { msg_send![self, GPUStartTime] }
+    }
+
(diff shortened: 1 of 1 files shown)


src/commandbuffer.rs
```diff
@@ -66,6 +66,14 @@ impl CommandBufferRef {
         }
     }
 
+    pub fn gpu_start_time(&self) -> f64 {
```
kvark, line 69: code is good, but could we not put it right between `label()` and `set_label()`? I'd suggest putting it right after `pop_debug_group()` instead

PR #331, opened 2024-07-16
Add Texture from IOSurface

fabiosky:
Added wrapper for:

`- (id<[MTLTexture](https://developer.apple.com/documentation/metal/mtltexture)>)newTextureWithDescriptor:([MTLTextureDescriptor](https://developer.apple.com/documentation/metal/mtltexturedescriptor) *)descriptor;`

https://developer.apple.com/documentation/metal/mtldevice/1433425-newtexturewithdescriptor

IOSurfaceRef is already defined in:
https://github.com/servo/core-foundation-rs/tree/main/io-surface

So importing as dependency.


MarijnS95:
The comment and link in your description are wrong, those are already there. In this PR you're adding:

https://developer.apple.com/documentation/metal/mtldevice/1433378-newtexturewithdescriptor

with signature:

```
- (id<MTLTexture>)newTextureWithDescriptor:(MTLTextureDescriptor *)descriptor 
                                 iosurface:(IOSurfaceRef)iosurface 
                                     plane:(NSUInteger)plane;`
```

fabiosky:
Renamed commit message and PR title/message.

src/device.rs
```diff
@@ -1985,6 +1985,19 @@ impl DeviceRef {
...
+    pub fn new_texture_with_descriptor_and_iosurface(
+        &self,
+        descriptor: &TextureDescriptorRef,
+        iosurface: io_surface::IOSurfaceRef,
+        plane: NSUInteger,
+    ) -> Texture {
+        unsafe {
+            msg_send![self, newTextureWithDescriptor: descriptor
+                                           iosurface: iosurface
+                                               plane: plane ]
```
MarijnS95, line 1997: Looks like this file lacks any consistent formatting, but it should probably be:

```suggestion
            msg_send![self, newTextureWithDescriptor:descriptor
                                           iosurface:iosurface
                                               plane:plane]
```

fabiosky: Resolved

Cargo.toml
```diff
@@ -35,6 +35,7 @@ block = "0.1.6"
 foreign-types = "0.5"
 dispatch = { version = "0.2", optional = true }
 paste = "1"
+io-surface = { git = "https://github.com/servo/core-foundation-rs" }
```
kvark, line 38: not that this is a rather confusing dependency. Ideally, it matches the crate name

kvark:
Also, ideally IOSurface is included in core-foundation-types...

PR #335, opened 2024-09-25
Add ComputePassDescriptorRef::set_dispatch_type

kvark:
See https://developer.apple.com/documentation/metal/mtlcomputepassdescriptor/3564435-dispatchtype?language=objc
Tested manually.


kvark:
Merging as small and non-risky

PR #337, opened 2024-09-25
Add AccelerationStructurePassDescriptor

kvark:
Also renames "accelerator_structure" module for consistency.
Tested manually in https://github.com/kvark/blade/pull/179


kvark:
@grovesNL could you take a look, please?

kvark:
@grovesNL pinging again

src/acceleration_structure_pass.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl AccelerationStructurePassSampleBufferAttachmentDescriptor {
+    pub fn new() -> Self {
+        let class = class!(MTLAccelerationStructurePassSampleBufferAttachmentDescriptor);
+        unsafe { msg_send![class, new] }
+    }
+}
+
+impl AccelerationStructurePassSampleBufferAttachmentDescriptorRef {
+    pub fn sample_buffer(&self) -> &CounterSampleBufferRef {
```
madsmtm, line 75: This is nullable
```suggestion
    pub fn sample_buffer(&self) -> Option<&CounterSampleBufferRef> {
```

madsmtm (approved):

src/acceleration_structure_pass.rs
kvark: Yes indeed, but that also applies to the other `sample_buffer` calls, so it's consistent.
I can switch them all to Options

kvark:
Thanks for anybody looking at this. I'd be comfortable waiting more, but I want to get Zed up to date, and it would better for it to link to this repo's metal instead of my fork. So merging this. Please shout out if you feel this to be wrong!

PR #343, opened 2025-01-02
Fix wrong pointer type of `static _dispatch_main_q` binding

MarijnS95:
While looking at a copy of this code (in another project that uses `objc2`, which does not yet define `dispatch` integration) this strange cast immediately stands out.  `_dispatch_main_q` is defined here as a `dispatch_queue_t`, the `queue` argument in `dispatch_data_create()` is also defined as a `dispatch_queue_t`, yet there is a "borrow" (get *pointer to*) operation before passing it into the function.

Instead, this `static` field is supposed to be defined as the `Object` itself, so that any user could take a _pointer to it_ if they wish or need it so.  The same is seen in [the `dispatch` crate]: `_dispatch_main_q` is defined as a `dispatch_object_s`, and `dispatch_queue_t` is a _typedef_ to a `*mut dispatch_object_s` to match the above convention.

Note that this is not a bugfix, but merely a readability improvement. Use of `mut` here is debatable.

[the `dispatch` crate]: https://github.com/SSheldon/rust-dispatch/blob/f540a2d8ccaebf0e87f5805033b9e287e8d01ba5/src/ffi.rs#L33

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -11,7 +11,12 @@ use block::Block;
 use log::warn;
 use objc::runtime::{NO, YES};
 
-use std::{ffi::CStr, os::raw::c_char, path::Path, ptr};
+use std::{
+    ffi::CStr,
+    os::raw::c_char,
+    path::Path,
+    ptr::{self, addr_of_mut},
+};
 
 /// Available on macOS 10.11+, iOS 8.0+, tvOS 9.0+
 ///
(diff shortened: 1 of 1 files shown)


MarijnS95:
CI correctly suggests usage of the [new `&raw` syntax](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#native-syntax-for-creating-a-raw-pointer), but that requires an MSRV bump to 1.82.

madsmtm (approved):
👍

Note that it might _technically_ be a soundness fix, since we now go from telling the compiler that `_dispatch_main_q` has the size of a pointer (and asserting the validity of that by creating a reference in `&_dispatch_main_q`), to using `Object` which is zero-sized, and hence acts as-if the object may have an arbitrary size (in a similar way that extern types do).

But I doubt it matters at all in practice (esp. since LLVM migrated to opaque pointers).

ErichDonGubler (approved):

ErichDonGubler:
@cwfitzgerald: Now that Firefox has an MSRV of Rust 1.82, I wouldn't object to a bump to Rust 1.82 (as follow-up) so we can use `&raw` here later. Thoughts?

cwfitzgerald:
Yeah that's fine.

PR #344, opened 2025-01-02
Fix all clippy lints

MarijnS95:
`metal-rs` seems to have gotten quite behind on `clippy` lints, making new contributions tedious thanks to almost 100 warnings in the CI (and in a local IDE).


src/sync.rs
```diff
@@ -90,6 +90,8 @@ foreign_obj_type! {
 }
 
 impl SharedEventListener {
+    /// # Safety
+    /// TODO
     pub unsafe fn from_queue_handle(queue: dispatch_queue_t) -> Self {
```
MarijnS95, line 95: Handle is a raw pointer, and probably needs to be valid.

src/counters.rs
```diff
@@ -71,7 +71,7 @@ impl CounterSampleBufferRef {
 
     pub fn resolve_counter_range(&self, range: crate::NSRange) -> Vec<NSUInteger> {
         let mut data = vec![0 as NSUInteger; range.length as usize];
-        let total_bytes = range.length * mem::size_of::<NSUInteger>() as u64;
+        let total_bytes = mem::size_of_val(data.as_slice()) as u64;
         unsafe {
             let ns_data: *mut crate::Object = msg_send![self, resolveCounterRange: range];
             let () = msg_send![ns_data, getBytes: data.as_mut_ptr() length: total_bytes];
```
MarijnS95, line 77: Might actually be nice to use `Vec::with_capacity()` + `spare_capacity_mut()`.

examples/texture/Cargo.toml
```diff
@@ -18,4 +18,4 @@ winit = "0.29"
 
 [dependencies.objc]
 version = "0.2.4"
-features = ["objc_exception"]
+# features = ["objc_exception"]
```
MarijnS95, line 21: Need to remove this: was needed to cross-compile the code from a Linux machine (and not have to get finicky with the **atrocity** of a Macbook keyboard).

madsmtm: See also https://github.com/gfx-rs/metal-rs/pull/303

MarijnS95: That only removes it from the root `Cargo.toml`, and mentions nothing about it still being used in the example :/

madsmtm: Yeah, I agree that this change should also be made in the examples, just wanted to give context for why it fails ;)

MarijnS95:
https://github.com/SSheldon/rust-objc/issues/125 still needs to be worked around. The linter error is a little vague:

```
warning: unexpected `cfg` condition value: `cargo-clippy`
```

it intended to say that `cfg(feature = "cargo-clippy")` isn't valid. We can't solve this with:

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(feature = "cargo-clippy")'] }
```

Because "it only supports simple identifiers". The only workaround is adding `cargo-clippy = []` to `[features]` or disabling this lint entirely.

ErichDonGubler:
@MarijnS95: I have some fixes and `clippy` version refreshing at https://github.com/erichdongubler/metal-rs/commit/e95365f9cc85, but can't push to this PR. Could you please either enable that, or change the tip to point to that instead?

Once that's done, happy to merge this.

ErichDonGubler (changes requested):
Getting this out of my review queue, but almost ready!

examples/raytracing/geometry.rs
```diff
@@ -416,14 +393,14 @@ impl Geometry for SphereGeometry {
...
-        descriptor.set_primitive_data_buffer(Some(&self.sphere_buffer.as_ref().unwrap()));
+        descriptor.set_primitive_data_buffer(Some(self.sphere_buffer.as_ref().unwrap()));
         descriptor.set_primitive_data_stride(size_of::<Sphere>() as NSUInteger);
         descriptor.set_primitive_data_element_size(size_of::<Sphere>() as NSUInteger);
         From::from(descriptor)
     }
 
     fn get_resources(&self) -> Vec<Resource> {
-        return vec![From::from(self.sphere_buffer.as_ref().unwrap().clone())];
+        vec![From::from(self.sphere_buffer.as_ref().unwrap().clone())]
```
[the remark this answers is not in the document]
ErichDonGubler: Hmm, this isn't worth holding up the review for. I'll do a separate PR for this.

examples/mesh-shader/main.rs
```diff
@@ -43,7 +42,7 @@ fn main() {
         if let Ok(RawWindowHandle::AppKit(rw)) = window.window_handle().map(|wh| wh.as_raw()) {
             let view = rw.ns_view.as_ptr() as cocoa_id;
             view.setWantsLayer(YES);
-            view.setLayer(mem::transmute(layer.as_ref()));
+            view.setLayer(<*mut _>::cast(layer.as_mut()));
```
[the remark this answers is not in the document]
ErichDonGubler: Resolved with my latest changes; once incorporated, I'll `Resolve` this.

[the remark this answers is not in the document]
MarijnS95: That won't work, because `layer.as_mut()` returns a reference/borrow which does not have a `.cast()` method to change the type of a raw pointer. That's why this relies on borrow->raw pointer coercion.

MarijnS95:
@ErichDonGubler while at it, I'd rather you put #343 _in_ your review queue so that we can get that "dependency" merged first, which improves one of the pointer casts already.

Secondly `metal-rs` _really_ needs to define an MSRV. This PR could benefit a lot from `&raw` pointers, relying on `size_ofxxx()` being in the prelude now, etc.

examples/texture/build.rs
```diff
@@ -53,21 +53,24 @@ fn generate_rust_types_from_shader_types() {
...
+            if old_hash == current_hash {
+                return;
+            }
         }
     }
 
     let bindings = bindgen::Builder::default()
         .header("shader_types/wrapper.h")
+        // Most types have a layout mismatch between the shader and the host?
+        .layout_tests(false)
```
MarijnS95, line 69: Layout tests for some shader primitive types are failing.

MarijnS95: Fixed by only generating bindings for the 3 types we need.

examples/texture/build.rs
```diff
@@ -53,21 +53,25 @@ fn generate_rust_types_from_shader_types() {
...
-    if let Some(old_hash) = read_cached_shader_types_hash() {
-        if old_hash == current_hash {
-            return;
+    let out = PathBuf::from(env::var("OUT_DIR").unwrap());
+    let out = out.join("shader_bindings.rs");
+    if out.exists() {
+        if let Some(old_hash) = read_cached_shader_types_hash() {
+            if old_hash == current_hash {
+                return;
+            }
```
MarijnS95, line 62: Any clue why this even exists? We already rely on `cargo` to track changes to shaders and aply rerun the compile when they did - why limit running bindgen in this incredibly silly way?

This above all prevents `bindgen` from rerunning when changing this build script, if the compiled shader hash didn't change 👎

MarijnS95: I've taken my due-diligence to just delete this flawed logic altogether. It serves no purpose other than to unnecessarily break workflows.

ErichDonGubler: I don't know why this check is here. I left it in because I don't allocate enough bandwidth for `metal` crate specifically to justify in-depth review.

`cargo check`, `cargo clippy` and `rust-analyzer` all have failures with this logic. It seems strictly unnecessary, so I'm in support of removing it.

examples/mesh-shader/main.rs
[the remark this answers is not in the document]
ErichDonGubler: Ah, I had misremembered the path here; one _can_ use `std::ptr::from_mut`, and I think that that's an easier path to understand.

I'll file a PR for after this as follow-up.

ErichDonGubler:
LGTM now; thank you for your patience here! 😅

examples/mesh-shader/main.rs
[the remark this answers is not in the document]
MarijnS95: I've seen that you wanted to use `ptr::from_xxx()` on your branch, but I don't necessarily understand that as a benefit when the Rust language provides coercion from borrows to raw pointers.

PR #350, opened 2025-03-08
Set and test MSRV at 1.82

MarijnS95:
Following https://github.com/gfx-rs/metal-rs/pull/343#issuecomment-2708428426

This also makes use of `size_of(_val)()` being imporced via the prelude, `&raw` pointers being available to - for example - save on casts from borrows to raw pointers, and C-string literals automatically appending NUL terminators to strings.


cwfitzgerald (approved):
Thanks!

PR #366, opened 2025-11-02
Provide gpu_resource_id for Metal Buffers

msvbg:
This mirrors what's available for e.g. textures and samplers, and is necessary to support storage buffers in binding arrays on Metal. There will be a corresponding wgpu PR that depends on this one.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -68,4 +68,8 @@ impl BufferRef {
     pub fn gpu_address(&self) -> u64 {
         unsafe { msg_send![self, gpuAddress] }
     }
+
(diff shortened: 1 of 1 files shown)


msvbg:
CI fails but I don't think it's my fault:

```
Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.82.0 (8f40fc59f 2024-08-21)).
  Consider trying a newer version of Cargo (this may require the nightly release).
```

cwfitzgerald:
Yeah CI hasn't passed in a while... this library is on serious life support

PR #382, opened 2026-04-10
Fix enum representation for ILP32 platforms (arm64_32)

matthargett:
All Metal framework enums use NS_ENUM(NSUInteger, ...) where NSUInteger is `unsigned long` — 8 bytes on LP64 (arm64) but 4 bytes on ILP32 (arm64_32, used by watchOS on Apple Watch Series 6–9 and SE2). Found this while making Multi-Draw Indirect in wgpu-native work on Apple Watch 6/SE2 (A13 CPU/GPU equivalent).

Using #[repr(u64)] forces all enums to 8 bytes regardless of platform, which corrupts the stack when passed via msg_send! on ILP32: the caller places 8 bytes where the ObjC runtime expects 4, misaligning every subsequent argument.

Fix: #[repr(u64)] -> #[repr(usize)], #[repr(i64)] -> #[repr(isize)]. This matches NSUInteger/NSInteger width on all Apple platforms.

Also fixes MTLDeviceLocation::Unspecified which used u64::MAX — this doesn't fit in a usize on ILP32. Changed to usize::MAX which correctly represents NSUIntegerMax on all platforms.

Zero behavioral change on 64-bit platforms (usize == u64).

Tested on:
- Apple Watch SE2 (arm64_32, S6/A13, watchOS 11.6)
- Apple Watch Series 10 (arm64, S9/A15, watchOS 26.4)

### DIFF
--- a/src/acceleration_structure.rs
+++ b/src/acceleration_structure.rs
@@ -19,7 +19,7 @@ bitflags::bitflags! {
 }
 
 /// See <https://developer.apple.com/documentation/metal/mtlaccelerationstructureinstancedescriptortype>
-#[repr(u64)]
+#[repr(usize)]
 #[allow(non_camel_case_types)]
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
 pub enum MTLAccelerationStructureInstanceDescriptorType {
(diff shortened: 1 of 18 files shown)


matthargett:
cc @cwfitzgerald

cwfitzgerald (approved):
Sorry about the long delay here!

PR #394, opened 2026-07-07
Use lockfileMaintence for non-breaking changes

MarijnS95:
Ported from https://github.com/gfx-rs/wgpu/pull/7629, to (hopefully) fix the completely broken titles that Renovate used to generate. CC @cwfitzgerald 

TODO: Check future changes like `minimumReleaseAge` and Rust toolchain upgrades from https://github.com/gfx-rs/wgpu/commits/trunk/renovate.json.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,18 +5,29 @@
   "prConcurrentLimit": 20,
   "prHourlyLimit": 200,
   "labels": ["dependencies"],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "recreateWhen": "always",
+    "rebaseStalePrs": true,
+    "branchTopic": "Cargo.lock update",
+    "commitMessageAction": "Update Cargo.lock",
+    "schedule": ["before 4am on monday"],
+    "prBodyDefinitions": {
+      "Change": "All locks refreshed"
+    }
+  },
   "packageRules": [
     {
       "matchUpdateTypes": ["patch"],
       "matchCurrentVersion": "<1.0.0",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -5,18 +5,29 @@
...
     },
     {
       "matchUpdateTypes": ["minor", "patch"],
       "matchCurrentVersion": ">=1.0.0",
-      "groupName": "Minor Updates",
-      "description": "Minor and patch updates to x.y.z crates are treated as compatible by cargo"
+      "enabled": false,
+      "description": "Minor and patch updates to x.y.z crates are compatible and handled by lockFileMaintenance"
     }
   ]
```
MarijnS95, line 32: Perhaps we could just merge these rules?

cwfitzgerald: I'm not positive we can, as the matchUpdateTypes is the same

cwfitzgerald (approved):
Thanks!

renovate.json
MarijnS95: ~Assuming `matchCurrentVersion` is optional, a single `"matchUpdateTypes": ["minor", "patch"],` should match both and disable the functionality entirely?~

~Could we just remove this entirely, assuming/hoping `lockfileMaintence` overrules any default rules?~

EDIT: Ah I might have just misread this as disabling all updates, but of course we still want those in `Cargo.toml` when the version is **not semver compatible** (i.e. outside of `Cargo.lock`). Seems like Renovate should have a mode for that if it also understands `lockfileMaintenance`?

